Issue #13, opened 2025-02-25
[Enhancement]: Metrics and logging


mrjvs:
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What enhancement would you like to see?

Juxt projects needs some logging and observability:
- Structured logging -- log in json or pretty based on config
- Log incoming requests
- Log unhandled exceptions
- Use prometheus client to add custom metrics (host on separate port for security)
- Both juxt projects should use the same logging system.

### Any other details to share? (OPTIONAL)

For metrics, there should be two new settings:
- `metrics.enabled: boolean` - If metrics should be enabled or not. Default to false
- `metrics.port: number` - What port should metrics be exposed on. Default to 9090.
- Metrics are on a separate port for security, we don't want to expose the metrics along with public endpoints

For logs, there is only one new setting:
- `logging.format: "json" | "pretty"` - What format to log in, Default to "pretty". Pretty is just nice readable logs with colors, JSON is just a line of JSON per log entry.
- Unhandled exceptions and incoming requests should be logged
- All console logs should go through the new logger, no `console.log` should be left out
- There shouldn't be any ignored errors using `catch {}`, at minimum it should log the error.
- Logs should only go to stdout/stderr, no need for log files or folders. (log directory can be removed after implementing)
- There are some popular log libraries out there. i.e. pino, winston

PR #105, opened 2025-04-28
Add Prometheus metrics clients

ashquarky:
Part of #13 

### Changes:

Adds [express-prom-bundle](https://www.npmjs.com/package/express-prom-bundle) and [prom-client](https://www.npmjs.com/package/prom-client) to expose various metrics for our Prometheus server to scrape. By default, the metrics are disabled, and once enabled they run on a separate port (default `localhost:9090/metrics`).

Generic stats like CPU time and garbage collector info is available, as well as response times and response codes per-endpoint.

Note I had to use an older express-prom-bundle v7.0.2 since the latest (v8.0.0) depends on Express 5 and fails typescript otherwise. If we update Express we can also update this.


apps/juxtaposition-ui/src/config.js
```diff
@@ -6,6 +6,8 @@ const schema = z.object({
 	logFolder: z.string().default(`${__dirname}/../logs`),
 	httpPort: z.coerce.number().default(8080),
 	httpCookieDomain: z.string().default('.pretendo.network'),
+	metricsEnabled: z.coerce.boolean().default(false),
```
mrjvs, line 9: unfortunately since its coming from environment variable `z.coerce.boolean()` doesnt work.
You can use `zodCoercedBoolean()` from `@neato/config` instead. See the docs here for details: https://neatojs.com/docs/config/api/utils#zod-coerced-boolean

also applies to miiverse-api/src/config.ts

mrjvs (changes requested):
Tested, it works nicely. Just one little review comment before it can get merged.

mrjvs (approved):

PR #106, opened 2025-04-28
Fix static asset compiling

mrjvs:
Every since the image history incident :tm: we haven't had proper compilation step for juxtaposition-ui.

This was due to a few issues:
- static assets (ejs, json, css, client side js) were not copied over to the dist folder
- static file routing was broken due to the __dirname being changed (bundling into one file)

Due to that, typescript files were made into JS files as a temporary measure, then we just ran `node src/server.js` directly.

---

These issues are now fixed:
- Static files are copied over using `esbuild-copy-files`
- Static file routing is fixed by not bundling into one file, but compiling every separately

The rest of the changes:
- Changed config.js back into typescript
- npm start script now uses the compiled codebase instead of src/server.js directly
- Added missing dash to the docker compose file

Note: I wasn't able to test it, it's hard to get juxt running. but I double checked the file structure and it *should* work.
Second note: We should purge cloudflare cache when deployed, so we can properly check if CSS files are fixed

### DIFF
--- a/.docker/docker-compose.yml
+++ b/.docker/docker-compose.yml
@@ -5,8 +5,8 @@ services:
     image: redis:7-alpine
     restart: unless-stopped
     ports:
-      - '6379:6379'
-    volumes: 
+      - "6379:6379"
+    volumes:
       - redis_data:/data
   mongo:
     image: mongo:8.0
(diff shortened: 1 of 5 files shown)


binaryoverload (approved):
LGTM

Issue #99, opened 2025-04-12
[Enhancement]: Remove, or at least improve, "duplicate post" checks


jonbarrow:
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What enhancement would you like to see?

Remove, or at least improve, the "duplicate post" checks used by both [`miiverse-api`](https://github.com/PretendoNetwork/juxtaposition/blob/be41afa9bc0e465411f8503d15a846606ec730ac/apps/miiverse-api/src/services/api/routes/posts.ts#L377-L391) and [`juxtaposition-ui`](https://github.com/PretendoNetwork/juxtaposition/blob/be41afa9bc0e465411f8503d15a846606ec730ac/apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.js#L301-L304). They are flawed and there are likely better ways to handle this

## Overview

Both servers implement a "duplicate post" check when making a new post. The server checks to see if a duplicate post is being made, and then rejects the request if one is found. The reason for this check, as stated by Jemma on Discord, is:

> Duplicate post checking is unfortunately a necessity. Regardless of rate limits, the browser has this issue where someone pressing the submit button repeatedly submits exponentially more and more requests. It appears to be a flaw with this version of webkit. I've considered moving the post sending logic over to js so I can prevent duplicate requests, but it's not something I've tried yet

However I don't fully agree with this, and the checks in place are very flawed

## Issues

Both servers implement `getDuplicatePosts` identically, so I'll only be referencing the one found in `miiverse-api` as it's typed. The current `getDuplicatePosts` is as follows:

https://github.com/PretendoNetwork/juxtaposition/blob/be41afa9bc0e465411f8503d15a846606ec730ac/apps/miiverse-api/src/database.ts#L100-L111

The only checks it performs are:

- Same sender
- Same text content
- Same painting
- Same screenshot
- Has no parent
- Is not removed

While at first glance this seems fine to check if a post is the same, it fails to take into account 2 very important pieces of information; the community and the creation time. These checks would prevent someone from making the "same" post in a different community, and regardless of how much time has passed. For example, someone could create a post saying "I love this game" in the Super Mario Maker community, and then 3 years later make a post in the Mario Kart 8 community also saying "I love this game", and the server would flag it as a "duplicate post"

[...]

PR #108, opened 2025-04-30
Replace logging with Pino-based logger & refactor API error handling for better error reporting

ashquarky:
Part of #13 

### Changes:

Adds a new logging framework (Pino) to juxtaposition-ui and miiverse-api. The primary benefit of this is JSON-based log output with lots of extra juicy context for analysis. Defaults to "pretty" output which is a rough match to the old logging format, but once JSON format is enabled things like the full request context are attached to each log message. (For example, if users reported a specific 115- error code, one could look up the frequency of its occurrence, which title ids and endpoints are involved, and when it started.)

Additionally, every try/catch and console call in both codebases was audited and replaced with a new logger call where suitable. Many errors which were previously ignored are now logged, so expect an uptick of error rates in production. We should refactor these areas to handle problems differently.

Also includes a refactor of miiverse-api's error handling to use a consistent style for returning errors to the client (necessary to include the extra API error code info in the log entries). This *does* change several error-case API responses so will need thorough testing to see if any API clients were expecting specific code returns - though most of these endpoints were previously returning a HTTP status code without an API error attached at all. Juxtaposition-ui was unchanged since all of its business logic will be refactored soon anyway. Maybe a good chance to deploy a beta instance to prod so the testers can try many different games?

(See also e.g. #99 where an endpoint was returning an error code that misled users on the client side - claiming bans when the true issue was a duplicate post. That is changed to a different error code now.)

Also includes a refactor of how PNID data is handled in miiverse-api. Since the auth middleware was grabbing the user info from the account server anyway, we can cache it on the Express request, letting the endpoint code use the cache rather than hitting gRPC a second time.

- [x] I have tested all of my changes. (**Basic functionality tested** - edge cases will need time to cook)


apps/miiverse-api/src/middleware/auth.ts
```diff
@@ -41,52 +40,50 @@ async function auth(request: express.Request, response: express.Response, next:
...
+		request.log.error(user, `Discovery data is missing for ${user.serverAccessLevel}`);
+		return serverError(response, ApiErrorCode.NO_DISCOVERY_DATA);
 	}
 
-	if (discovery.status !== 0) {
-		return serverError(response, discovery);
+	if (discovery.status > 0 && discovery.status <= 7) {
+		return badRequest(response, discovery.status);
+	} else if (discovery.status !== 0) {
+		request.log.error(discovery, `Discovery for ${user.serverAccessLevel} has unexpected status`);
```
mrjvs, line 82: I think it's better if we don't log normal objects as if they are errors. Instead we can add the invalid param to the log message:
```ts
request.log.error(`Discovery status ${discovery.status} for ${user.serverAccessLevel} is unexpected`)
```

apps/miiverse-api/src/services/api/routes/communities.ts
```diff
@@ -315,7 +284,7 @@ router.post('/', multer().none(), async function (request: express.Request, resp
 
 	const ownedSubcommunityCount = await Community.countDocuments(ownedQuery);
 	if (ownedSubcommunityCount >= 4) {
-		return respondCommunityError(response, 401, 911);
+		return badRequest(response, ApiErrorCode.CREATE_TOO_MANY_COMMUNITIES, 403);
```
mrjvs, line 287: I see a lot of changes in error codes. The most common ones:
- 401 unauthorized changed to 403 forbidden
- 422 changed to more semantically correct status codes

Are these changes all intentional? Are we sure this won't cause issues with consoles, do they care about the status codes?

apps/miiverse-api/src/services/discovery/routes/discovery.ts
```diff
@@ -11,92 +10,36 @@ const router = express.Router();
...
-			break;
-		default:
-			message = 'SERVER_ERROR';
-			errorCode = 15;
-			response.type('application/xml');
-			break;
+	if (discovery.status > 0 && discovery.status <= 7) {
+		return badRequest(response, discovery.status);
+	} else if (discovery.status !== 0) {
+		request.log.error(discovery, `Discovery for ${request.user?.serverAccessLevel} has unexpected status`);
```
mrjvs, line 29: Same comment as before

apps/juxtaposition-ui/src/server.js
```diff
@@ -111,4 +111,9 @@ async function main() {
 	}
 }
 
-main().catch(console.error);
+process.on('uncaughtException', (err) => {
+	logger.fatal(err, 'Uncaught exception');
+	process.exit(1);
+});
+
+main();
```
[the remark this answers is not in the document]
mrjvs: Same thing with miiverse-api

mrjvs (changes requested):
looks good :)

apps/juxtaposition-ui/src/server.js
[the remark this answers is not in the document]
ashquarky: Docs are a little unclear but I think it does. See the `origin` argument here: https://nodejs.org/api/process.html#event-uncaughtexception

apps/miiverse-api/src/services/api/routes/communities.ts
ashquarky: They are intentional (e.g. 403 is because we're already past the authentication middleware by now, so Unauthorized is the wrong code) but I'm not sure how the clients react. We can at least expect different pop-up error codes: https://github.com/PretendoNetwork/miiverse-api/wiki/Miiverse-Error-Codes
I smoketested it and things still worked, but this is the sort of thing where edge cases could come up (maybe a client is expecting a 404 on a certain community)

apps/miiverse-api/src/config.ts
```diff
@@ -3,7 +3,9 @@ import { z } from 'zod';
 
 // schema is mapped later to nested object to keep env vars consistent with other projects
 const schema = z.object({
-	logFolder: z.string().default(`${__dirname}/../logs`),
+	logFormat: z.enum(['json', 'pretty']).default('pretty'),
+	logLevel: z.enum(['error', 'warn', 'info', 'debug', 'trace']).default('info'),
+	logRedact: zodCoercedBoolean().default(true),
```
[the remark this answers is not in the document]
ashquarky: Sure, good idea

apps/miiverse-api/src/services/api/routes/communities.ts
ashquarky: I've done additional testing with Nintendo Land (apparently one of the pickier titles) and I wasn't able to get it to do anything weird on the error paths; though it's still pretty limited and challenging to hit all the paths

mrjvs: Good that you gave it some extra tests. For the rest we'll just test in production, we can do a rollback if it's *too* unstable.

mrjvs (approved):

PR #109, opened 2025-05-07
Path aliases for juxtaposition-ui project

mrjvs:
changes:
- Fixed some linting errors
- Updated vscode config to conflict with less extensions
- Updated the juxtaposition-ui compilation step to allow for path aliases
- Updated the juxtaposition-ui project files to use path aliases

This change brings it closer together with standards set in miiverse-api (which already uses path aliases)


ashquarky:
🔥

PR #114, opened 2025-05-08
juxtaposition-ui: Check for expired tokens in frontend

ashquarky:
The API server checks device type *and* token expiry - bring juxtaposition-ui in line with that

### Changes:

### DIFF
--- a/apps/juxtaposition-ui/src/util.js
+++ b/apps/juxtaposition-ui/src/util.js
@@ -128,6 +128,11 @@ function processServiceToken(encryptedToken) {
 			return null;
 		}
 
+		// * Check if the token is expired
+		if (token.expire_time < Date.now()) {
+			return null;
+		}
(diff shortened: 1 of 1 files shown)


DaniElectra:
The issue with this is that the token currently isn't renewed, which may cause issues on the consoles. Ideally we want to add token renewal before adding the expiry check

DaniElectra:
I have checked on the code of Miiverse on Ghidra and it seems that at least on the Miiverse Posting app sets a cache of 12 hours for the token. Since the Miiverse applet uses the same client ID, I assume the time would be the same there, and thus the service token is expected to expire at that time

ashquarky:
Confirmed the account server always uses 1 hour expiry unconditionally, which is problematic here

ashquarky:
Superseded by #131

Issue #102, opened 2025-04-26
[Feature]: Moderation - Interface for Logged Actions


Joshua8600:
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What feature do you want to see added?

Suggested in the April 26, 2025 mod meeting is an interface to be able to view the logging that already exists in the database:
 -past actions on a user
 -moderator-removed posts
 -which moderator did what actions (bans, limit from posting, removals)
 -submitted reports (in the case of ignored/false reports)
 -anything else useful for moderation (ask other mods for their opinions)

### Why do you want to have this feature?

Basically, Juxt Moderation is in an unsustainable state. Having to manually log everything, when we get anywhere from 10-20 offending reports *per day*, is exhausting. It leads to missed logs, incorrect logs, and miscommunication between moderators.

ashquarky:
>  which moderator did what actions (bans, limit from posting, removals)

The other features are added, but this one is blocked by #15.
We can also use this issue to discuss improvements to the other features (I have already had pagination get requested)

PR #116, opened 2025-05-11
Moderation improvements and fixes

ashquarky:
### Changes:

Various moderation-related changes to try and streamline Juxtaposition moderation.
- Adds two report lists (reports against the user, and reports by the user); along with a list of recently deleted posts and their deletion reasons, to the existing admin/users page. Partial response to #102 
- Enables display of deleted posts and banned user profiles to moderator accounts. #41 
- Uses Moment to give fuzzy ban dates ("banned for 1 day", "banned for a week") etc. #36 

The code isn't the finest (especially the EJS) but I figured it's not the end of the world for internal tooling. I've posted screenshots of some of the changes in the Discord staff channels.

This PR allows Juxt mods to view information they previously did not have access to:
- Deleted posts, even those deleted by the user
- Moderation actions and notes from other moderators when deleting posts or processing reports

Jon has already approved all the issues referenced, so I know it's okay from the leadership side, but I want to make sure this is okay from the mod team side as well, so I'll keep this on draft until the headmods give the nod. We can start with code review while we wait, since any issues they bring up will just be "revert the bad part".

### DIFF
--- a/apps/juxtaposition-ui/src/util.js
+++ b/apps/juxtaposition-ui/src/util.js
@@ -246,6 +246,23 @@ function updateCommunityHash(community) {
 	communityMap.set(community.olive_community_id, community.name);
 }
 
+function getReasonMap() {
+	return [
+		'Spoiler',
+		'Personal Information',
+		'Violent Content',
+		'Inappropriate/Harmful Conduct',
+		'Hateful/Bullying',
+		'Advertising',
+		'Sexually Explicit',
+		'Piracy',
+		'Inappropriate Behavior in Game',
+		'Other',
+		'Missing Images; Reach out to Jemma with post link to fix',
+		'Unknown'
+	];
+}
+
 async function resizeImage(file, width, height) {
 	return new Promise(function (resolve) {
 		const image = Buffer.from(file, 'base64');
@@ -515,6 +532,7 @@ module.exports = {
 	refreshCache,
 	setName,
 	updateCommunityHash,
+	getReasonMap,
 	resizeImage,
 	getTGAFromPNG,
 	createBMPTgaBuffer,
--- a/apps/juxtaposition-ui/src/util.js
+++ b/apps/juxtaposition-ui/src/util.js
@@ -258,8 +258,7 @@ function getReasonMap() {
 		'Piracy',
 		'Inappropriate Behavior in Game',
 		'Other',
-		'Missing Images; Reach out to Jemma with post link to fix',
-		'Unknown'
+		'Missing Images; Reach out to Jemma with post link to fix'
 	];
 }
 
--- a/apps/juxtaposition-ui/src/webfiles/web/moderate_user.ejs
+++ b/apps/juxtaposition-ui/src/webfiles/web/moderate_user.ejs
@@ -161,7 +161,7 @@
 									</span>
 									<span class="text">
 										<h4>
-											<%= reasonMap[report.reason] %>
+											<%= reasonMap[report.reason] ?? "Unknown" %>
 										</h4>
 										<p>
 											<%= report.message %>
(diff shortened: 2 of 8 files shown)


apps/juxtaposition-ui/src/webfiles/web/moderate_user.ejs
```diff
@@ -128,45 +128,231 @@
...
+											<% break;
+											case 9: %>
+											Other
+											<% break;
+											case 10: %>
+											Missing Images; Reach out to Jemma with post link to fix
+											<% break;
+											default: %>
+											Unknown
+											<% break; } %>
```
binaryoverload, line 200: I would possibly do this mapping in the code that passes in the `reports` variable - imo there should be minimal logic inside the template for clarity and any data mapping is done before the data is given to the template 

Thoughts?

binaryoverload (changes requested):
Looks good! Just a few comments :)

ashquarky:
Received the nod from shog and sky, mod team is on board. Will address review comments in a bit :)

apps/juxtaposition-ui/src/webfiles/web/moderate_user.ejs
ashquarky: Yeah fair enough. I had copied this from the existing code but it's pretty rough.
There are other places where mapping is done explicitly, just through objects instead of spelt out like this - `communitiesMap`, `userMap`, `postMap` etc. Changing this design is probably a whole different refactor?

binaryoverload: Maybe we have a `rulesMap` which you give the reason ID to and it has the rule name as the value in the map? That would be a relatively small change, keeping in scope with this PR

apps/juxtaposition-ui/src/util.js
```diff
@@ -246,6 +246,23 @@ function updateCommunityHash(community) {
...
+		'Inappropriate/Harmful Conduct',
+		'Hateful/Bullying',
+		'Advertising',
+		'Sexually Explicit',
+		'Piracy',
+		'Inappropriate Behavior in Game',
+		'Other',
+		'Missing Images; Reach out to Jemma with post link to fix',
+		'Unknown'
+	];
```
binaryoverload, line 263: This won't handle Unknown correctly, as the unknown should be the "default"

You'll want to remove unknown from the array, and use a null coalescing operator to return unknown `reasonMap[report.reason] ?? "Unknown"`

binaryoverload (changes requested):
The changes look good, just one for the reasons :)

binaryoverload (dismissed):

apps/juxtaposition-ui/src/webfiles/web/moderate_user.ejs
```diff
@@ -128,45 +128,159 @@
...
 			</div>
 			<div class="mt-5 text-center">
 				<button class="btn btn-primary profile-button" type="button" onclick="savePNID()">Save User</button>
 			</div>
 		</div>
-    </div>
+		<div class="mt-5 text-center">
+			<h4 class="text-right">Recently Reported Posts (limit 5)</h4>
+		</div>
+		<ul class="list-content-with-icon-and-text arrow-list" id="news-list-content">
```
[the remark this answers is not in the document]
binaryoverload: Yeah this is a good point @ashquarky - the IDs should either be manually changed or maybe just removed if they're not needed?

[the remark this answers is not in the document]
ashquarky: They do affect the CSS slightly, but I don't want to get into that right now, so I've just removed them. The padding is different now but it still looks fine

mrjvs (approved):
Will and I share a brain cell, so im approving it in his name :P

Nice job, im sure mods will be hyped. Ill look into a doing a release this evening

PR #118, opened 2025-05-12
Logging improvements

### LINKED ISSUE #117: [Bug]: X-Real-Ip or X-Forwarded-For does not properly resolve into the remoteAddress log
### Checked Existing

### What happened?

In new JSON formatted logs, the `req.remoteAdress` field does not show the proper IP address thats resolved from the IP proxy headers: X-Real-Ip or X-Forwarded-For.

I see two possible causes:
1. remoteAddress does not account for the headers at all. so we should log `request.ip` instead
2. The proxy headers are not properly extracted, which would then cause issues with ratelimiting.

It needs some investigation to see what is actually going on.

Show the correct IP address

### Steps to reproduce?

1. make a request to juxt-ui using header: `X-Forwarded-For: ?.?.?.?`
2. Enable JSON formatted logs
3. `req.remoteAddress` is not set to the IP address of the header

_No response_

ashquarky:
Resolves #117

### Changes:

Uses express request.ips instead of connection so that we get X-Forwarded-For/proxy working. Also some other changes to reduce bloat in the JSON logger.


apps/miiverse-api/src/loggerHttp.ts
```diff
@@ -25,15 +26,20 @@ export const loggerHttp = pinoHttp({
...
 
 			// Decode param pack if we have it
-			if ('x-nintendo-parampack' in req.headers) {
-				req.param_pack = decodeParamPack(req.headers['x-nintendo-parampack']);
+			const paramPack = raw.headers['x-nintendo-parampack'];
+			if (typeof paramPack === 'string') {
+				req.param_pack = decodeParamPack(paramPack);
 			}
 
+			req.remoteAddress = raw.ip ?? req.remoteAddress;
```
mrjvs, line 41: I don't think miiverse-api has `trust proxy` set to true.

Maybe it's out of scope for this PR, but it means `raw.ip` won't represent proxy headers like we would expect them to.

we add a new setting`serve.trustProxy: boolean` (default false) (on both juxt-ui and miiverse-api) and then conditionally set the "trust proxy" if enabled. Or we can add it in a different PR later, since we're only really using it for logs rn.

jonbarrow: > raw.ip ?? req.remoteAddress

This check seems to not do anything, `raw.ip` should always return a value? `??` only returns the right hand if the left hand is `null` or `undefined`, which Express should never let happen? Checking the source for both Express and its dependencies, I don't think `req.remoteAddress` will ever get used in this case?

- `req.raw` is set by Pino as the [incoming request object](https://github.com/pinojs/pino-std-serializers/blob/a7cfba4de6a6f9334de81984aa84d3c21c562b9d/lib/req.js#L92) (unless you're using hapi)
- Since we don't use hapi, the `raw` variable is just the exact same as `req` (so that assignment could also be removed tbh)
- Express defines [`req.ip` as a getter](https://github.com/expressjs/express/blob/9784321e89b7d32aaff481aa24a9e8cccef4b101/lib/request.js#L327-L330), so it will always exist. It returns the result of `proxyaddr`, which is what we'd need to worry about not returning a value. But it seems like it always does?
- `proxyaddr` comes from [`proxy-addr`](https://github.com/jshttp/proxy-addr)
- `proxyaddr` internally calls `alladdrs` and returns the last item of the array
- `alladdrs` immediately calls `forwarded` from [`forwarded`](https://github.com/jshttp/forwarded) before doing anything
- `forwarded` returns [an array of addresses](https://github.com/jshttp/forwarded/blob/a2c8c86c99b150fa536c3f37c0c167fa3be7744f/index.js#L24-L36) which will always contain [the socket address](https://github.com/jshttp/forwarded/blob/a2c8c86c99b150fa536c3f37c0c167fa3be7744f/index.js#L46-L50)

Unless I'm missing something?

> we add a new setting`serve.trustProxy: boolean` (default false) (on both juxt-ui and miiverse-api) and then conditionally set the "trust proxy" if enabled.

Is there a reason to not leave `trust proxy` just always enabled? The UI app already enables it at all times, it's not configurable. And based on what I saw in how `req.ip` works it doesn't look like it would do any harm? If the setting is enabled but there is no `X-Forwarded-For` header (like when there is no reverse proxy) then it just returns the socket address (effectively the same as having the setting disabled)

I know that this would mean that the server would *always* trust the `X-Forwarded-For` header, so even if the server is not behind a reverse proxy you could send a `X-Forwarded-For` header anyway, but is that actually an issue? The Express docs have a warning that the field is inherently untrustworthy no matter what, so I'm not sure what new issues it would introduce by just leaving it on all the time. And 99% of people would be running this behind a reverse proxy of *some* sort in a real deployment, so it would be configured on in most cases anyway?

I’m not against making it a setting, to be clear, just trying to understand the logic and from what I can see it would just add an extra step, but I could be missing something

mrjvs: There are scenarios where software is ran without a reverse proxy. It's a little unusual but not rare.

Trusting proxy headers is inherently risky, so it should be opt-in. Just like express is.

mrjvs: Also just checked. Jon is right, `req.ip` is marked as required, so the fallback to remoteAddress is likely unneeded.

So to summarize the changes for the review comment:
- add a `server.trustProxy: boolean` (default false) config var. Either now, or in a separate PR at a later time.
- Remove the fallback to `req.remoteAddress`, express guarantees the `req.ip` field exists

ashquarky: TypeScript says no: 
image

I guess this is a bug in the Express types.

ashquarky:
I implemented the proxy setting, with some extra code to not just handle true/false but also allow setting trusted IPs and hop counts, as per the Express docs: https://expressjs.com/en/guide/behind-proxies.html

I would recommend for prod this be set to `loopback, uniquelocal` or `10.0.0.0/8`; though you could also put the IP of the nginx instance here.

apps/juxtaposition-ui/src/config.ts
```diff
@@ -8,6 +8,7 @@ const schema = z.object({
 	logSensitive: zodCoercedBoolean().default(false),
 	httpPort: z.coerce.number().default(8080),
 	httpCookieDomain: z.string().default('.pretendo.network'),
+	httpTrustProxy: z.union([zodStrictBoolean(), z.coerce.number(), z.string()]).default(false),
```
mrjvs, line 11: The new config option supports a lot of cases - a lot more complicated then a boolean.

I think it's wise to add a comment explaining the basic inputs (cidr & boolean) and a link to https://expressjs.com/en/guide/behind-proxies.html for details.

(same for miiverse-api)

mrjvs (changes requested):

apps/miiverse-api/src/loggerHttp.ts
mrjvs: > I guess this is a bug in the Express types.

Weird, I thought I checked it.

Anyway that solves it, just keep the `??`

PR #121, opened 2025-05-23
web: Pack client JS and add hamburger menu to web frontend

### LINKED ISSUE #104: [Feature]: Deletion without a report
### Checked Existing

Moderators should have the ability to remove a post without having to report it beforehand.

It would make moderation of individual posts and mass removal of spam a lot easier for moderators.

_No response_

ashquarky:
Resolves #104

### Changes:

- Runs client JS/CSS through the existing ESBuild pipeline, and remove the corresponding pre-minified files. This enables packing/bundling and some light Babel-y stuff for CTR; though only minification is in use right now.
  - Each build targets a different platform (Node, modern, Chrome20, IE7) so four tsup builds are needed. The watch feature started four Juxt instances because of this? Switch to nodemon so we get one instance.
- Tidy up the post_template partial (and lightly touch some of the other partials). Refactor out icons into their own files (still inlined by EJS)
- Refactor the report modal. Previously, you had to go into a post page, and some posts had a flag icon, but others had a large "report" button, and all of those would click another hidden button in JavaScript, which triggered PJAX, which... Just add reporting to a hamburger menu and have it hide/unhide a proper modal.
- Since we have a hamburger menu, add post deletion (#104) for moderators.


apps/juxtaposition-ui/package.json
```diff
@@ -5,7 +5,8 @@
   "license": "AGPL-3.0-only",
   "main": "./dist/server.js",
   "scripts": {
-    "dev": "tsup --watch --onSuccess \"node --enable-source-maps dist/server.js\"",
+    "dev": "nodemon -x '/bin/sh -c' -- 'tsup && node --enable-source-maps dist/server.js'",
```
binaryoverload, line 8: This unfortunately won't be compatible with Windows (or Mac? 🤔)

I think the best way would be to have:
`dev` - runs `nodemon --exec 'npm run dev:buildrun'`
`dev:buildrun` - uses [npm-run-all](https://www.npmjs.com/package/npm-run-all) to run two scripts sequentially using `run-s build start`

apps/juxtaposition-ui/src/webfiles/web/js/post.js
```diff
@@ -0,0 +1,11 @@
+export function deletePost(id, reason) {
+	if (!id) {
+		return;
+	}
+
+	fetch(`/posts/${id}?reason=${reason ?? ''}`, {
+		method: 'DELETE'
+	})
+		.then(res => res.text())
+		.then(_res => location.reload());
```
binaryoverload, line 10: Is there a reason that `.text()` is called and then the result isn't used?

binaryoverload (changes requested):
Looks pretty good! Just a couple of comments:

binaryoverload (changes requested):
Thanks for making those changes - looks good!

Just needs a `npm run lint:fix` :)

ashquarky:
> Thanks for making those changes - looks good!
> 
> Just needs a `npm run lint:fix` :)

That's a vendored library :( Not sure why it's started failing now when it didn't before. `lint:fix` bails out so I'll have to work something else out

binaryoverload (approved):

Issue #112, opened 2025-05-07
[Enhancement]: Moving datasource from juxtaposition-ui to miiverse-api


mrjvs:
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What enhancement would you like to see?

Right now, juxtaposition-ui and miiverse-api are both connecting to mongodb and modifying the data directly from their own connections. We want to move to have only one source of truth, which will be miiverse-api.

The JSON API will live on `/api/v1` and it will be separate express router from the olive API also living in miiverse-api. 

I've been in a voice call with @CaramelKat and @ashquarky and we've drafted the following endpoints: (this is a flexible list, if something doesn't sit right when implementing then just change it)
```ts
GET /communities/popular
GET /communities // paged filter:sort=createdAt filter:parent=ID
GET /communities/tid:{{title_id}}
GET /communities/{{id}}
POST /communities/{{id}}/posts
GET /users/{{id}}
GET /users/@me/profile
GET /users/@me/profile/settings
PATCH /users/@me/profile/settings
GET /users/@me/followers // paged - who is following the user
GET /users/@me/following // paged - who the user following
POST /users/{{id}}followers/@me
DELETE /users/{{id}}/followers/@me
GET /communities/{{id}}/posts // paged sort:empathy sort:createdAt filter:verified
POST /communities/{{id}}/follower/@me
DELETE /communities/{{id}}/follower/@me
GET /activity-feed // paged
GET /posts // paged filter:topic_tag filter:user_id filter:empathy_by
GET /posts/{{id}}
DELETE /posts/{{id}}
GET /posts/{{id}}/replies // paged
POST /posts/{{id}}/replies
GET /posts/{{id}}/empathies
POST /posts/{{id}}/empathies
GET /users/@me/conversations // paged
GET /conversations/{{id}}
POST /conversations/{{id}}/messages
GET /conversations/{{id}}/messages // paged
GET /users/@me/notifications // paged
GET /users/@me/notification-stats // conversations + notification count
GET /users/@me/friend-requests // paged
POST /users

// + admin tools
```

### Any other details to share? (OPTIONAL)

The core idea was already in #10 but that was closed prematurely (whoops), This is a more fleshed out version of that main issue.

mrjvs:
All of the details of this issue have been invalidated. Corporate wants to change to grpc

PR #127, opened 2025-06-08
grpc: Move /posts/<id> to backend

ashquarky:
Part of #112 
Need to submit this before it gets unreviewable

### Changes:

This PR adds the extra infrastructure needed for miiverse-api to act as datasource (authentication, error handling, type conversions, validation etc.) and then moves the "post by id" endpoint (and only this endpoint) to the new backend. A few calls to the old database exist in the frontend since they rely on the Post being a Mongoose object for further DB calls - these will be ported once those further calls are in the API.

- The Endpoint: Just sends the document right out of MongoDB for now.
- Validation: On the frontend side, that document gets passed through a Zod schema. The one I've written (Post) is *descriptive*, not *prescriptive*, made in consultation with the prod database so all the jank of the current DB schema is preserved. This will be useful to make sure changes we make to miiverse-api don't break the frontend.
- Authentication: Either the OAuth token (web), service token (console) or nothing (guest web) are included with requests to the backend as headers. It has middlewares to process these tokens, and also has per-endpoint middlewares to select between guest, user, moderator access. On the frontend side, a "UserTokens" object keeps the current state of things cached and can be blindly passed back into the API, so this doesn't make much mess in the business logic.
- Error handling (API): Errors can be thrown deliberately to get specific HTTP codes (Unauthorized, Forbidden etc.), or can just be general JavaScript errors. I know throwing is a different design to what I had refactored into the Nintendo API, but I think this is nicer again. Maybe I'll re-refactor the Nintendo API. 
- Error handling (frontend): HTTP codes from the backend are thrown and allowed to bubble. The error handlers and rendering are tweaked to work better (instead of sometimes spewing a stacktrace), include the request ID, etc. A special case is included for when the user needs to re-log-in; the frontend is extremely permissive about caching and not re-checking tokens, see also #114 - I'm expecting this might be a problem in prod

I know this is an absolute boatload of infrastructure for a very minimalist endpoint (and even it's missing some things, like the moderator/removed posts override) but hopefully it will serve as a foundation for us to work on top of.

### DIFF
--- a/apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.js
+++ b/apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.js
@@ -12,6 +12,7 @@ const upload = multer({ dest: 'uploads/' });
 const redis = require('@/redisCache');
 const { config } = require('@/config');
 const router = express.Router();
+const { apiFetchUser } = require('@/fetch');
 
 const postLimit = rateLimit({
 	windowMs: 15 * 1000, // 30 seconds
@@ -108,7 +109,7 @@ router.post('/new', postLimit, upload.none(), async function (req, res) {
 router.get('/:post_id', async function (req, res) {
 	const userSettings = await database.getUserSettings(req.pid);
 	const userContent = await database.getUserContent(req.pid);
-	let post = await database.getPostByID(req.params.post_id.toString());
+	let post = await apiFetchUser(req, `/api/v1/posts/${req.params.post_id}`);
 	if (post === null) {
 		return res.redirect('/404');
 	}
--- a/apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.js
+++ b/apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.js
@@ -114,7 +114,7 @@ router.get('/:post_id', async function (req, res) {
 		return res.redirect('/404');
 	}
 	if (post.parent) {
-		post = await database.getPostByID(post.parent);
+		post = await apiFetchUser(req, `/api/v1/posts/${post.parent}`);
 		if (post === null) {
 			return res.sendStatus(404);
 		}
--- a/apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.js
+++ b/apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.js
@@ -12,7 +12,7 @@ const upload = multer({ dest: 'uploads/' });
 const redis = require('@/redisCache');
 const { config } = require('@/config');
 const router = express.Router();
-const { apiFetchUser } = require('@/fetch');
+const backend = require('@/backend');
 
 const postLimit = rateLimit({
 	windowMs: 15 * 1000, // 30 seconds
@@ -42,7 +42,7 @@ const yeahLimit = rateLimit({
 });
 
 router.get('/:post_id/oembed.json', async function (req, res) {
-	const post = await database.getPostByID(req.params.post_id.toString());
+	const post = await backend.getPostById(req.tokens, req.params.post_id);
 	const doc = {
 		author_name: post.screen_name,
 		author_url: 'https://juxt.pretendo.network/users/show?pid=' + post.pid
@@ -109,16 +109,11 @@ router.post('/new', postLimit, upload.none(), async function (req, res) {
 router.get('/:post_id', async function (req, res) {
 	const userSettings = await database.getUserSettings(req.pid);
 	const userContent = await database.getUserContent(req.pid);
-	let post = await apiFetchUser(req, `/api/v1/posts/${req.params.post_id}`);
-	if (post === null) {
-		return res.redirect('/404');
-	}
+
+	const post = await backend.getPostById(req.tokens, req.params.post_id);
 	if (post.parent) {
-		post = await apiFetchUser(req, `/api/v1/posts/${post.parent}`);
-		if (post === null) {
-			return res.sendStatus(404);
-		}
-		return res.redirect(`/posts/${post.id}`);
+		const parent = await backend.getPostById(req.tokens, post.parent);
+		return res.redirect(`/posts/${parent.id}`);
 	}
 	const community = await database.getCommunityByID(post.community_id);
 	const communityMap = await util.getCommunityHash();
@@ -166,7 +161,7 @@ router.post('/:post_id/new', postLimit, upload.none(), async function (req, res)
 
 router.post('/:post_id/report', upload.none(), async function (req, res) {
 	const { reason, message, post_id } = req.body;
-	const post = await database.getPostByID(post_id);
+	const post = await backend.getPostByID(req.tokens, post_id);
 	if (!reason || !post_id || !post) {
 		return res.redirect('/404');
 	}
--- a/apps/juxtaposition-ui/src/webfiles/web/error.ejs
+++ b/apps/juxtaposition-ui/src/webfiles/web/error.ejs
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
     <!-- general -->
-    <title>Juxtaposition Log In</title>
+    <title>Juxt - Error</title>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <link rel="manifest" href="/web/manifest.json">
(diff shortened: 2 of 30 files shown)


apps/miiverse-api/src/services/internal/middleware/authentication.ts
```diff
@@ -0,0 +1,63 @@
...
+async function consoleAuth(serviceToken: string): Promise<AccountData> {
+	const pid = getPIDFromServiceToken(serviceToken);
+	if (pid === 0) {
+		throw new errors.unauthorized('Invalid service token');
+	}
+
+	const pnid = await getUserAccountData(pid);
+	// If the user has a valid token for an unknown PID, just let the exception bubble
+
+	const settings = await getUserSettings(pid) ?? undefined; // Null doesn't play nice with TS ?
```
mrjvs, line 42: > Null doesn't play nice with TS

You write the type to only allow undefined. If you want to use `null` then you have to add that like its null: `settings: ... | null`

apps/miiverse-api/src/services/internal/middleware/check-user-account.ts
```diff
@@ -0,0 +1,23 @@
...
+	}
+
+	if (account.pnid.accessLevel < 0) {
+		throw new errors.unauthorized('Account has been banned');
+	}
+
+	// TODO account settings, temp bans, etc.
+
+	return next();
+}
```
mrjvs, line 21: Maybe we should more action driven middlewares like `canUserWrite`.
Cuz banned users can still read posts, right?

apps/miiverse-api/src/services/internal/middleware/authentication.ts
```diff
@@ -0,0 +1,63 @@
...
+	})).pid;
+	// Ask the "backdoor" API, just use the above as a glorified token decryption.
+	const pnid = await getUserAccountData(pid);
+
+	const settings = await getUserSettings(pid) ?? undefined;
+
+	return { pnid, settings };
+}
+
+export default auth;
```
mrjvs, line 63: Let's avoid default exports in any new code we write. It makes importing files way more difficult since it all needs to be done manually by devs. Named exports can be magically imported using editor features.

apps/miiverse-api/src/services/internal/middleware/check-user-account.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export default checkUserAccount;
```
mrjvs, line 23: Use a named export, see other comment

mrjvs (changes requested):

apps/miiverse-api/src/services/internal/routes/posts.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const postsRouter = express.Router();
+
+postsRouter.get('/:post_id', authGuest, handle(async ({ req }) => {
+	const post = await getPostByID(req.params.post_id);
+	if (!post || post.removed) {
+		throw new errors.notFound('Post not found');
+	}
+
+	return post;
```
mrjvs, line 15: To ensure a strict data contract from the API. I usually have a mapping function `mapPost(dbPost)`.
It basically just does this:
```ts
export type PostDto = {
  id: string;
  foo: ...;
  bar: ...;
}

export function mapPost(post: TypeGoesHere): PostDto {
  return {
    id: post.id,
    foo: ...,
    bar: ...
  }
}
```

This has a few benefits:
- It allows you to make your API nicer before it's sent to the client. Like formatting dates or structuring types better.
- It ensures that devs knows what's supposed to be returned. If it doesn't match the output type def then it's unintentional.
- It minimizes the amount of variants of the same data we have, you're less likely to add an extra field for one specific endpoint or change the logic for one endpoint. You'd just use `mapPost` and the types are unified.

In this specific context:
- We can fix the problem of some things being & undefined in the same dataset. The frontend only gets the one variant cuz we can map it here.
- We can't change too much of the data model, since the frontend still relies on the same structure. But it will provide a good base for refactoring later

mrjvs (changes requested):

apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.js
```diff
@@ -108,16 +109,11 @@ router.post('/new', postLimit, upload.none(), async function (req, res) {
...
 	if (post.parent) {
-		post = await database.getPostByID(post.parent);
-		if (post === null) {
-			return res.sendStatus(404);
-		}
-		return res.redirect(`/posts/${post.id}`);
+		const parent = await backend.getPostById(req.tokens, post.parent);
+		return res.redirect(`/posts/${parent.id}`);
 	}
 	const community = await database.getCommunityByID(post.community_id);
```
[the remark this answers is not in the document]
ashquarky: It does - 404 will throw a FetchError (inside apiFetch) which bubbles up to the generic Express error handler and renders this page: 
image
(This is a slightly different page to the Express 404 handler.. so some tweaking is needed no matter what. I could special-case the code, which is already happening for 401 to trigger a re-login.)

I did think about using `null` here, I'm just on the fence about having to put if null return 404 everywhere in the business logic. OTOH, maybe there are places where we actually want to have "if object doesn't exist, do something different"?

apps/juxtaposition-ui/src/webfiles/web/error.ejs
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
     <!-- general -->
-    <title>Juxtaposition Log In</title>
+    <title>Juxt - Error</title>
```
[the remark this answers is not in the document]
ashquarky: Should I also update the other page titles that use Juxt?

apps/miiverse-api/src/services/internal/middleware/check-user-account.ts
ashquarky: I'm pretty sure banned users can't even log in and have to use the anonymous access. I do still have a bit of action-driven stuff on the routes itself:
```js
postsRouter.get('/:post_id', authGuest, //...
```
This would allow guest (and I guess I could special-case banned too) reading of posts; but we'd have something else like:
```js
postsRouter.post('/:post_id/empathy', authUser, //...
```
...so only logged-in users can Yeah.

apps/miiverse-api/src/services/internal/routes/posts.ts
ashquarky: Alright, I don't mind that idea. It would allow fixing some of the more egregious typing errors (like boolNumber) without having a corresponding DB migration (until we're ready)

ashquarky:
Note to self: Need to check Juxt bans on the API too (not just PNID bans)

apps/juxtaposition-ui/src/webfiles/web/error.ejs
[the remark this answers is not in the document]
mrjvs: With the context that every other page uses juxt, lets just use Juxt.

apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.js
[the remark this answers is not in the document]
mrjvs: I think 404's specifically need to manually handled. Normal errors can be bubbled up since we don't really need to handle them on a case by case basis, but 404's and validation errors will need to be done explicitly (imo)

apps/miiverse-api/src/services/internal/middleware/check-user-account.ts
mrjvs: How about these middleware?
```ts
router.use(populateAuth); // token handling
router.use(authAccess); // prevent unauthorized users from accessing

router.get('/:id', guards.guest, ...);
router.get('/:id', guards.user, ...);
router.get('/:id', guards.mod, ...);
```

apps/miiverse-api/src/services/internal/routes/posts.ts
```diff
@@ -0,0 +1,17 @@
...
+import { guards } from '@/services/internal/middleware/guards';
+import { mapPost } from '@/services/internal/contract/post';
+
+export const postsRouter = express.Router();
+
+postsRouter.get('/posts/:post_id', guards.guest, handle(async ({ req }) => {
+	const post = await getPostByID(req.params.post_id);
+	if (!post || post.removed) {
+		throw new errors.notFound('Post not found');
+	}
```
mrjvs, line 14: It just crossed my mind, moderators should be able to view removed posts. right?

This check will need a bypass for mods.

mrjvs (changes requested):

mrjvs (approved):

mrjvs:
lgtm, is this ready for merge?

ashquarky:
Token expiry issue https://github.com/PretendoNetwork/account/issues/184

mrjvs:
This is waiting on https://github.com/PretendoNetwork/account/issues/184 to be resolved, can be merged when that's done

ashquarky:
Account server side is good to go, now just need our side (which is present in #138) and this can land

PR #129, opened 2025-06-09
Chore: cleanup posting logic

mrjvs:
changes:
- Cleanup posting logic

There was a TODO to clean up the logic, so here it is!
The original logic did have a few bugs, biggest one being that `access_level` field doesn't exist.

@ashquarky can you go through and run some tests, make sure all the checks are correct?
@CaramelKat it would also be nice if you'd take a look at it, extra pair of eyes is always good.

### DIFF
--- a/apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.js
+++ b/apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.js
@@ -190,6 +190,32 @@ router.post('/:post_id/report', upload.none(), async function (req, res) {
 	return res.redirect(`/posts/${post.id}`);
 });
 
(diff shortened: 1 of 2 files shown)


CaramelKat (approved):
As far as I can tell this is pretty solid. Trying it locally with a couple options seemed to be fine, but I've not hit every edge case (another point to us having automated tests lol). I'd like to get Luca's opinion as well just to be sure but I think we're good :+1:

ashquarky:
Looks like a developer account can't post to announcements with this setup (it passes the access level requirement, but fails on `isPulicPostableCommunity`).
I am. actually unsure what the intended behaviour is here? I have posted to announcements with my dev acccount before.

PR #131, opened 2025-06-17
feat(auth): Implement new 'issue time' field on service tokens

ashquarky:
Resolves #XXX

### Changes:

Implements the proposed NNAS service token format from https://github.com/PretendoNetwork/account/issues/184#issuecomment-2977601545 . Service tokens now have an issue time instead of an expiry time, and Juxt is responsible for expiring old tokens. This allows us to extend the validity period (at least 12 hours is needed, I've gone for 24 to be safe). 

Having correct token expiry behaviour will allow us to phone back to the account server much more often (currently we can't because they expire too soon) and properly enforce bans and other token requirements. Ideally, I would like to validate tokens on every request, perhaps with a short (1-10min) cookie cache for reads.

This does increase our dependence on decrypting service tokens, but that was already the case due to the token type check, which needs account-server-side changes. When those happen, the issue time field can be added there too.

Unable to test due to requiring paired changes on the account server side. This *should* be alright to deploy - it'll just make old-style tokens last much longer - but it'd be better to have the account server fixed at the same time.

### DIFF
--- a/apps/juxtaposition-ui/src/util.js
+++ b/apps/juxtaposition-ui/src/util.js
@@ -121,10 +121,15 @@ function processServiceToken(encryptedToken) {
 	try {
 		const B64token = Buffer.from(encryptedToken, 'base64');
 		const decryptedToken = this.decryptToken(B64token);
-		const token = this.unpackToken(decryptedToken);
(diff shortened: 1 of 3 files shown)


ashquarky:
Will roll into #138

Issue #12, opened 2025-02-25
[Enhancement]: JSX based templating


mrjvs:
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What enhancement would you like to see?

The current Juxtaposition-UI project uses EJS for templating.

Unfortunately we don't get typesafety or autocomplete, this causes bugs.

JSX is typesafe and has autocompletion, it's a good choice for a templating engine.

Note this will only be used on the backend for SSR, not on the frontend (as the consoles are too slow)

PR #132, opened 2025-06-19
Add JSX view engine

mrjvs:
changes:
 - Add JSX renderer util to response object
 - Port common partials needed for most views to JSX (web only)
 - Port `messages.ejs` to JSX (web only)
 - Update all (web) endpoints that use `messages.ejs` to use the new JSX based view

This partially contains #12 
 
---
 
Since it may be useful for future porting, here is roughly the steps I took to port EJS to JSX:
1. Make a new component for the new view
2. Copy paste the old EJS code into new view
3. Replace EJS conditionals and loops with JSX (conditional: `{ cond ? < /> : < />}`, loops: `{ arr.map(v => (< />)}`)
4. Close all selfclosing tags (`` -> ``)
5. Replace `class` attributes with `className`
6. Replace `style=""` attributes with `style={{}}` (jsx takes an object and not a string)
7. Replace dynamic text content with JSX style: `<% hi %>` -> `{hi}`
8. Replace dynamic attributes with JSX version: `src="<%cdnUrl%>/test.png"` -> `src={\`${cdnUrl}/test.png\``
9. Update all standard HTML attributes (stroke-width) to camelcase (strokeWidth) - exclude custom attributes that aren't standard
10. Raw style tag content and script tag content can be set using a new util: `InlineScript` and `InlineStyle`

Some more caveats:
- There are no globals from `res.locals`. So I made a context object to passs around and put all of that in.
- Partials don't exist, simply make a new component
- While this PR does include React, we're only using the JSX tools. This system only generates HTML with 0 javascript, so state or reactivity or contexts, it all doesn't work. It can only generate static HTML.


mrjvs:
I updated the code to render all the different consoles/directories. Also tested all 3 of them, they seem to be working

ashquarky:
Also worth noting that our ESLint config doesn't seem to cover TSX/JSX files. I guess https://www.npmjs.com/package/@pretendonetwork/eslint-config needs updating to match the new file extensions

mrjvs:
@ashquarky I have addressed all review comments. Also added made an eslint config PR but we don't have to wait for it imo.

binaryoverload (changes requested):
https://github.com/PretendoNetwork/common-configs/pull/3 is merged, update pls :)

mrjvs:
Nothing to worry about since that has to do with reactivity engine, which we don't use and can't use.

Although it's a bit sloppy to leave it there indefinitely, I'll fix it next time I work on Jsx related tasks

PR #135, opened 2025-06-21
Fixed personal feed showing up in global feed

mrjvs:
changes:
- Removed duplicate endpoint, the fixed version already existed, but never triggered because of the duplicate


ashquarky (approved):
I get the feeling this code kinda bitrotted as it wasn't in use? Works now tho

PR #161, opened 2025-08-20
feat(ui): Image processing improvements

### LINKED ISSUE #33: [Bug]: Banner resolution for 3DS is wrong.
### Checked Existing

- [X] I have checked the repository for duplicate issues.

### What happened?

When trying to add new communities, the community tab is reporting 400x220, but that's not the correct resolution.
Original Miiverse banners on 3DS used 400x168 images, but even if we decide to not use original Miiverse banners, it's still the wrong resolution overall, since Juxtaposition has still the community title at the bottom of the top screen, similar to Miiverse.

This bug can be circumvented by adding dummy pixels to 400x168 images to make them 400x220, and allowing setting banners on 3DS as intended. Setting 400x168 images currently just "zooms" the image.

Image resolution standard being 400x168 for Nintendo 3DS.

### Steps to reproduce?

1. If your PNID is set to accessLevel = 3 (Developer), go to Moderation
2. Click Communities
3. Add a new community, as you can see, the resolution reported is 400x220! (Keep in mind, the 3DS screen is 400x240, so this should be already clear it's not the right one) 

This bug has been encountered when @MatthewL246 was testing and setting up new communities.

The current tab is this:
image

ashquarky:
Resolves #33 

### Changes:
Miiverse painting of a fox saying awoo

This PR has numerous image and image-processing themed improvements.
- Use ImageMagick (wasm) for all processing. This simplifies our dependency tree dramatically and removes quite a few NPM audit warnings.
- Decode and compress **paintings** on the server. Previously the TGA sent by the console was converted to a PNG, but the TGA was also saved and served unmodified to API clients. Now this TGA is decoded and re-encoded (losslessly) to enforce limits on width, height, colour, etc. On the PNG rendering side, using bilevel colour and tuned compression parameters routinely gets paintings under a kilobyte (the above sample is 796 bytes).
- Decode and compress **screenshots** on the server. The Wii U uploads these as JPEGs, but again we now re-encode these to enforce limits on filetype, colour, resolution. Also, we now generate a 320px "thumbnail" version for display on 3DS, which will hopefully help with its RAM shortage. Store aspect ratio information about uploaded screenshots so we can support "other" sizes later. :)
- Decode and compress **headers** on the server. Despite the .png extension on the CDN, these could be any file type. Now, compress them as progressive JPEGs to help with bandwidth usage and to improve perceived load times. Admin panel users **must** upload real PNGs now as source images. Handle both 220px (Juxt native) and 168px (Miiverse native) 3DS headers, padding the 168px assets with white pixels. 3DS headers uploaded after this PR will also cause the title area to become transparent, allowing custom designs to be put there. Otherwise, the aforementioned white pixels just show through.
- Replace the default header image with a 4-bit colour version. The visual difference is very minor and it's something like 10KiB of base64 blob saved. At some point I would like to consider if we have too many inline images on 3DS and if it's causing rendering delays.
- Decode and compress **icons** on the server. Not much to say here, just PNG compression parameters and stripping of extra data. Minor improvement.
- Gracefully handle Friends server failures and outages
- Developers using the access_token cookie trick correctly have their tokens passed to the internal API now

### DIFF
--- a/apps/miiverse-api/src/server.ts
+++ b/apps/miiverse-api/src/server.ts
@@ -10,6 +10,7 @@ import api from '@/services/api';
 import { healthzRouter } from '@/services/healthz';
 import { config } from '@/config';
 import { setupGrpc } from '@/services/internal/server';
+import { initImageProcessing } from '@/images';
 import { ApiErrorCode, badRequest, serverError } from './errors';
 
 const { http: { port }, metrics: { enabled: metricsEnabled, port: metricsPort } } = config;
@@ -79,6 +80,7 @@ async function main(): Promise<void> {
 	logger.info('Starting server');
 
 	await connectDatabase();
+	await initImageProcessing();
 
 	app.listen(port, '0.0.0.0', () => {
 		logger.info(`Server started on port ${port}`);


apps/miiverse-api/src/types/mongoose/post.ts
```diff
@@ -15,7 +15,9 @@ export interface IPost {
 
 	painting?: string; // base64, can be empty or undefined
 	screenshot?: string; // url fragment (leading /), can be empty or undefined
+	screenshot_thumb?: string; // url fragment (leading /), can be empty or undefined
 	screenshot_length?: number;
+	screenshot_aspect?: string; // '4:3' '5:4' '16:9'
```
mrjvs, line 20: Miiverse-API also has some image processing in `services/api/routes/posts.ts` and `services/api/routes/communities.ts`. Should we also include the changes from the PR there?

apps/miiverse-api/src/images.ts
```diff
@@ -0,0 +1,351 @@
...
+		wup
+	};
+}
+
+export async function initImageProcessing(): Promise<void> {
+	const wasmPath = new URL(import.meta.resolve('@imagemagick/magick-wasm/magick.wasm'));
+	const wasm = await readFile(wasmPath);
+	await initializeImageMagick(wasm);
+	logger.success('Started up image processing engine');
+}
```
mrjvs, line 351: I think you forgot to call this method in the entrypoint

mrjvs (changes requested):
One small comment and it needs a quick `npm run lint:fix` on the api project.

Then it should be ready to go!

PR #163, opened 2025-08-23
Migrate messages router and views to typescript 

mrjvs:
Changes:
- Migrate all EJS views in messages router to typescript JSX (That includes new_post and report_modal partial)
- Move all responses to the end of the route handlers
- Port messages route to typescript
- Added a util to get authenticated requests, used for typing
- Ported conversation.js to typescript, so I can add method types to it
- Improved docker compose slightly: added postgres + bind on 0.0.0.0, so docker compose works on windows
- Stole some code from #161 -> web tokens now work on console endpoints
- Fixed JSX problems with events, now you use `evt-EVENTNAME` and jsx renderer will set it correctly. We may want to consider using preact, as that doesn't rewrite any attributes.
- Fixed JSX compile warnings, it prefers `defaultValue` and `defaultChecked` for input elements if they aren't managed by react reactivity

Special changes in views:
- Added a loop for empathies instead of hardcoding all the inputs
- Button on web: `header-post-button` had a href="#". Which is unused and also invalid html that JSX didnt like
- Had to add a data-alert-text attribute to one text area (PortalNewPostView Line 101). So I don't have to generate JS with string templating.

Special changes when porting to typescript:
- Request didnt have the right type, so I made a `getAuthedRequest` which checks if middleware have ran and then casts the request object.
- `params.post_id` was checked for, but it could never exist, so I just removed that code
- `generatePostUID` was bugged if it encounters an inuse ID due to missing parameter
- mii's are technically nullable, so I added an assertion before creation of a post

Resolves a part of #12


apps/juxtaposition-ui/src/models/conversation.ts
ashquarky, line 1: This is different to how it's handled in miiverse-api, which uses an `IConversation` type and then:
```ts
export const Conversation = model<IConversation, ConversationModel>('Conversation', ConversationSchema);
```
https://github.com/PretendoNetwork/juxtaposition/blob/dev/apps/miiverse-api/src/models/conversation.ts
https://github.com/PretendoNetwork/juxtaposition/blob/dev/apps/miiverse-api/src/types/mongoose/conversation.ts

Your approach does look a little cleaner; but I think it's more valuable to have consistency given this codebases' history of the two apps disagreeing on the database schema. Maybe converting miiverse-api to this new style is a good idea?

apps/juxtaposition-ui/src/types/middleware.ts
```diff
@@ -0,0 +1,21 @@
...
+	pid: number;
+	tokens: {
+		serviceToken?: string;
+		oauthToken?: string;
+	};
+	paramPackData: null | ParamPack;
+};
+
+export function getAuthedRequest<TReq extends Request = Request>(req: TReq): AuthRequest<TReq> {
+	if (!(req as any).user) {
```
ashquarky, line 17: I think this boxes us in a little, since if the middleware were to be converted next, it wouldn't be able to set `req.user`, since that's not in the base Request type.

I think it would be nicer to extend the base Request type so our custom bits are available everywhere. We already do this with the `directory` field: https://github.com/PretendoNetwork/juxtaposition/blob/dev/apps/juxtaposition-ui/src/types/express.d.ts#L6-L8
Also miiverse-api, where it's used for the same data: https://github.com/PretendoNetwork/juxtaposition/blob/dev/apps/miiverse-api/src/types/express.d.ts

Eventually it would be good to introduce something like the AccountData struct from `internal` everywhere (would simplify guest access a lot), but I recognise that's out of scope for this PR.

apps/juxtaposition-ui/src/types/middleware.ts
```diff
@@ -0,0 +1,21 @@
+import type { Request } from 'express';
+import type { GetUserDataResponse as ApiGetUserDataResponse } from '@pretendonetwork/grpc/api/get_user_data_rpc';
+import type { GetUserDataResponse as AccountGetUserDataResponse } from '@pretendonetwork/grpc/account/get_user_data_rpc';
+import type { ParamPack } from '@/types/common/param-pack';
+
+export type AuthRequest<TReq extends Request = Request> = TReq & {
+	user: ApiGetUserDataResponse | AccountGetUserDataResponse;
```
ashquarky, line 7: Let's change:
https://github.com/PretendoNetwork/juxtaposition/blob/0bc28757d6a16e318b3cf2b90aded3069ed34682/apps/juxtaposition-ui/src/middleware/webAuth.js#L14-L16
to:
```ts
try {
    request.pid = (await getUserDataFromToken(request.cookies.access_token)).pid;
    request.user = await getUserAccountData(request.pid);
```
Similarly for consoleAuth:
https://github.com/PretendoNetwork/juxtaposition/blob/0bc28757d6a16e318b3cf2b90aded3069ed34682/apps/juxtaposition-ui/src/middleware/consoleAuth.js#L27-L31
```ts
// Developer accounts may also use an OAuth token for console frontends
const user = await getUserDataFromToken(request.cookies.access_token);
if (user.accessLevel === 3) {
    request.user = await getUserAccountData(user.pid);
    request.pid = user.pid;
```

This will make it consistently use `AccountGetUserDataResponse`, which is the better one with more information (and has extra data needed for #134)

apps/juxtaposition-ui/src/services/juxt-web/views/portal/messageThread.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+	return (
+		<div
+			id={`message-${msg.id}`}
+			className={cx('post scroll', {
+				'my-post': msg.pid === props.ctx.pid,
+				'other-post': msg.pid !== props.ctx.pid
+			})}
+		>
+			<a href={`/users/show?pid=${msg.pid}`} data-pjax="#body" className="scroll-focus mii-icon-container">
```
ashquarky, line 26: Not essential for this PR, but we should think about a helper class using URLSearchParams so we don't muck up the encoding in more complex cases (it does URL escaping and stuff)

ashquarky (commented):
JSX stuff looks nice

apps/juxtaposition-ui/src/models/conversation.ts
mrjvs: I took the miiverse-api one and decided to make it simpler. but you're right that desyncs the two projects.

What do you think, should I take the one from miiverse-api or update miiverse-api with my one?

apps/juxtaposition-ui/src/types/middleware.ts
mrjvs: We discussed this in DM's and concluded that we can improve it later. I added a TODO with a little detail.

apps/juxtaposition-ui/src/types/middleware.ts
mrjvs: Done!

apps/juxtaposition-ui/src/services/juxt-web/views/portal/messageThread.tsx
mrjvs: Done!

ashquarky:
We can merge this and then I'll rebase #161 on top of it (since it tweaked the markup in a now-deleted EJS file I think)

PR #164, opened 2025-08-28
Migrate feed router and views to typescript 

mrjvs:
Changes:
- Migrate all EJS views in feed router to typescript JSX (That includes feed.ejs, posts_list.ejs, post_template.ejs)
- Add util to preprocess requests (auth context and zod schemas for query and body)
- Port feed router to typescript

Special changes in views:
- there was `post.url` handling on all platforms, this is now removed as the database does not include any url fields. Replaced with a TODO
- There was handling for `locals.yeah` on ctr, which doesn't seem to exist anymore. It was either `post.yeahs` or the requesting users yeahs. Replaced with a TODO
- Added some TODO's for handling moderation views, it's only properly handled on web
- There were multiple unclosed spans in `post_template.ejs` for portal near the topic_tag handling. Updated so it isn't broken.

Todos in scope for this PR:
- There is a second `<body>` element in the wiiu views for `feed.ejs`. I am unsure what to do with it
- SVG icons are not included properly
- Testing, nothing has been tested

Resolves a part of #12


apps/juxtaposition-ui/src/services/juxt-web/routes/console/feed.tsx
```diff
@@ -0,0 +1,167 @@
...
+	}
+	const posts = await database.getNewsFeed(userContent, config.postLimit);
+
+	const bundle = {
+		posts,
+		open: true,
+		communityMap,
+		userContent,
+		nextLink: `/feed/more?offset=${posts.length}&pjax=true`
+	};
```
ashquarky, line 37: Is there any point having "bundle" here instead of just local variables, given you pull all the members back out again during rendering?
Is it referenced somehow behind the scenes?

(e.g. communityMap seems unused here, the post listings get this out of props.ctx)

apps/juxtaposition-ui/src/services/juxt-web/routes/console/feed.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+export const feedRouter = express.Router();
+
+feedRouter.get('/', async function (req, res) {
+	const { auth, query } = parseReq(req, {
+		query: z.object({
+			pjax: z.string().optional()
+		})
+	});
+	const userContent = await database.getUserContent(auth().pid);
```
ashquarky, line 24: I don't think I really like this pattern.. it would be very easy to accidentally not call `auth` when making a handler that happens to not reference userContent. Having it inside a parameter makes it look less important even though this is a load-bearing permissions check

apps/juxtaposition-ui/src/services/juxt-web/routes/console/feed.tsx
```diff
@@ -0,0 +1,167 @@
...
+feedRouter.get('/more', async function (req, res) {
+	const { auth, query } = parseReq(req, {
+		query: z.object({
+			pjax: z.string().optional(),
+			offset: z.coerce.number().nonnegative().default(0)
+		})
+	});
+	const userContent = await database.getUserContent(auth().pid);
+	if (!userContent) {
+		throw new Error('Usercontent not found on new page');
```
ashquarky, line 105: Other handlers redirect to /404 here

apps/juxtaposition-ui/src/services/juxt-web/routes/console/messages.tsx
```diff
@@ -29,28 +29,30 @@ messagesRouter.get('/', async function (rawReq, res) {
 	});
 });
 
-messagesRouter.post('/new', async function (rawReq, res) {
-	const req = getAuthedRequest(rawReq);
+messagesRouter.post('/new', async function (req, res) {
+	// TODO add body validation
+	const { auth } = parseReq(req);
+	const authCtx = auth();
```
ashquarky, line 35: This is probably nicer and should be done everywhere rather than inlining the auth call

apps/juxtaposition-ui/src/services/juxt-web/routes/routeUtils.ts
```diff
@@ -0,0 +1,82 @@
...
+	};
+	paramPackData: null | ParamPack;
+};
+
+export type AuthContext = {
+	user: AccountGetUserDataResponse;
+	pid: number;
+	tokens: {
+		serviceToken?: string;
+		oauthToken?: string;
```
ashquarky, line 23: We have a type for this already https://github.com/PretendoNetwork/juxtaposition/blob/dev/apps/juxtaposition-ui/src/fetch.ts#L57 - we can move it into types/ and use it everywhere

apps/juxtaposition-ui/src/services/juxt-web/views/ctr/post.tsx
```diff
@@ -0,0 +1,112 @@
+import cx from 'classnames';
+import moment from 'moment';
+import { utils } from '@/services/juxt-web/views/utils';
+import type { ReactNode } from 'react';
+import type { PostViewProps } from '@/services/juxt-web/views/web/post';
+
+export function CtrPostView(props: PostViewProps): ReactNode {
+	const post = props.post;
+	const hasYeahed = post.yeahs && post.yeahs.indexOf(props.ctx.pid) !== -1;
+	// TODO implement moderator removed post logic
```
ashquarky, line 10: Probably can't release if we have regressed moderation functionality

apps/juxtaposition-ui/src/services/juxt-web/views/web/icons.tsx
```diff
@@ -0,0 +1,27 @@
+import heartIcon from '../../../../webfiles/web/partials/assets/heart_icon.svg?raw';
+import replyIcon from '../../../../webfiles/web/partials/assets/reply_icon.svg?raw';
+import menuIcon from '../../../../webfiles/web/partials/assets/menu_icon.svg?raw';
+import flagIcon from '../../../../webfiles/web/partials/assets/flag_icon.svg?raw';
+import binIcon from '../../../../webfiles/web/partials/assets/bin_icon.svg?raw';
+import shareIcon from '../../../../webfiles/web/partials/assets/share_icon.svg?raw';
```
ashquarky, line 6: :crying_cat_face:

ashquarky (changes requested):
Coming together well overall, stuff like `cx` makes me happy

apps/juxtaposition-ui/src/services/juxt-web/views/ctr/post.tsx
ashquarky: Oh I had misread the PR intro, this isn't a regression - no worries then

apps/juxtaposition-ui/src/services/juxt-web/routes/console/feed.tsx
mrjvs: `bundle` was so I didn't modify too much of the original route handlers. But I do agree that we can replace it by now.

apps/juxtaposition-ui/src/services/juxt-web/routes/console/feed.tsx
mrjvs: The authentication middleware is still applied, it won't get to the route without being authenticated. My intent was for `auth()` to be basically just a getter with an extra assertion so typescript is happy.

apps/juxtaposition-ui/src/services/juxt-web/views/web/icons.tsx
mrjvs: Unfortunately the esbuild plugin im using doesn't support path aliases. So I'd say we keep it like this until we find a better plugin (or method)

apps/juxtaposition-ui/src/services/juxt-web/routes/console/feed.tsx
ashquarky: Ok, fair enough

apps/juxtaposition-ui/src/services/juxt-web/routes/console/feed.tsx
```diff
@@ -0,0 +1,136 @@
...
+		web: <WebPersonalFeedView ctx={buildContext(res)} title={title} nextLink={nextLink} posts={posts} userContent={userContent} />,
+		portal: <PortalPersonalFeedView ctx={buildContext(res)} title={title} nextLink={nextLink} posts={posts} userContent={userContent} />,
+		ctr: <CtrPersonalFeedView ctx={buildContext(res)} title={title} nextLink={nextLink} posts={posts} userContent={userContent} />
+	});
+});
+
+feedRouter.get('/all', async function (req, res) {
+	const { auth, query } = parseReq(req, {
+		query: z.object({
+			pjax: z.string().optional()
```
ashquarky: I suppose `pjax` could be a stringbool now we're on Zod 4

ashquarky (approved):
LGTM. I like this design a lot better than the AccountData stuff now that I've seen what that entails

Approving with an optional minor change

apps/juxtaposition-ui/src/services/juxt-web/views/web/post.tsx
```diff
@@ -73,7 +73,7 @@ export function WebPostView(props: PostViewProps): ReactNode {
 				style={{
 					display: post.is_spoiler ? 'none' : undefined // Will be removed by spoiler-overlay onclick
 				}}
-				evt-click={`location.href='/posts/${post.id}`}
+				evt-click={`location.href='/posts/${post.id}'`}
```
ashquarky: honestly why isn't this just an <a> tag. how did we get here
don't change this, it's how the original is, more venting that we should do that redesign and redo the HTML markup at the same time

PR #168, opened 2025-09-22
feat(ctr): Bundling, optimise assets, page weight/memory improvements

ashquarky:
Resolves #XXX

### Changes:

- Bundle ctr code. This allows for tree shaking etc. Including Pjax with the bundler also means one less http request during page load.
- Only deliver debug.js to non-console UAs. It had a check in it to do nothing on console, so why download the 25KiB?
- Re-created various images and theme assets to compress better. Stylesheet goes from 82KiB to 37KiB. Also included the raw assets in the repo, in case we swap to a better system of inlining.
- Calls cave.requestGc after PJAX page loads. This is supposed to help alleviate RAM issues on O3DS a bit, but more probably needs doing.

### DIFF
--- a/apps/juxtaposition-ui/src/services/juxt-web/views/ctr/root.tsx
+++ b/apps/juxtaposition-ui/src/services/juxt-web/views/ctr/root.tsx
@@ -6,11 +6,12 @@ export type DefaultHeadProps = {
 };
 
 function DefaultHead(props: DefaultHeadProps): ReactNode {
+	const addDebugJs = !props.ctx.uaIsConsole; // Only serve debug js to non-console browsers
 	return (
 		<>
 			<link rel="stylesheet" type="text/css" href="/css/juxt.css" />
 			{/* Debug allows non-console browsers to have some amount of the cave API. */}
-			{props.ctx.uaIsConsole ?? true ? <></> : <script src="/js/debug.global.js"></script>}
+			{addDebugJs ? <script src="/js/debug.global.js"></script> : null}
 			<script src="/js/juxt.global.js"></script>
 		</>
 	);


apps/juxtaposition-ui/src/services/juxt-web/views/ctr/root.tsx
```diff
@@ -1,17 +1,23 @@
...
 		<>
 			<link rel="stylesheet" type="text/css" href="/css/juxt.css" />
-			<script src="/js/debug.js"></script>
-			<script src="/js/pjax.js"></script>
-			<script src="/js/juxt.js"></script>
+			{/* Debug allows non-console browsers to have some amount of the cave API. */}
+			{props.ctx.uaIsConsole ?? true ? <></> : <script src="/js/debug.global.js"></script>}
+			<script src="/js/juxt.global.js"></script>
 		</>
 	);
```
mrjvs, line 16: I had trouble understanding the logic, This should be the same outcome?
```suggestion
const addDebug = !props.ctx.uaIsConsole; // Add to non-console browsers
return (
<>
<link rel="stylesheet" type="text/css" href="/css/juxt.css" />
			{/* Debug allows non-console browsers to have some amount of the cave API. */}
			{addDebug ? <script src="/js/debug.global.js"></script> : null}
			<script src="/js/juxt.global.js"></script>
</>
);
```

apps/juxtaposition-ui/src/webfiles/ctr/js/debug.js
```diff
@@ -1,17 +1,3 @@
...
-		if (evt.message) { // Chrome sometimes provides this
-			alert('error: ' + evt.message + ' at linenumber: ' + evt.lineno + ' of file: ' + evt.filename);
-		} else {
-			if ((evt.srcElement || evt.target).toString().indexOf('HTMLImageElement') !== -1) {
-				return;
-			}
-			alert('error: ' + evt.type + ' from element: ' + (evt.srcElement || evt.target));
-		}
-	}
-}
```
mrjvs, line 14: What did this do?

apps/juxtaposition-ui/src/services/juxt-web/views/ctr/root.tsx
```diff
@@ -1,17 +1,23 @@
...
+function DefaultHead(props: DefaultHeadProps): ReactNode {
 	return (
 		<>
 			<link rel="stylesheet" type="text/css" href="/css/juxt.css" />
-			<script src="/js/debug.js"></script>
-			<script src="/js/pjax.js"></script>
-			<script src="/js/juxt.js"></script>
+			{/* Debug allows non-console browsers to have some amount of the cave API. */}
+			{props.ctx.uaIsConsole ?? true ? <></> : <script src="/js/debug.global.js"></script>}
+			<script src="/js/juxt.global.js"></script>
```
mrjvs, line 14: Where did `.global` come from, I don't see it in the compiling code anywhere

mrjvs (changes requested):

apps/juxtaposition-ui/src/webfiles/ctr/js/debug.js
ashquarky: On 3DS consoles (cave not undefined) it would show an alert dialog for JavaScript errors. I had never seen nor heard of anyone seeing these alerts, and the plan was to stop shipping this file to 3DSes anyway, so I yoinked it.
We could add it back into juxt.js if it serves a useful debugging format.

apps/juxtaposition-ui/src/services/juxt-web/views/ctr/root.tsx
ashquarky: That seems more sensible yeah

PR #196, opened 2025-10-02
Implement account deletion GRPC method

mrjvs:
Changes:
- Cleaned up internal GRPC protobufs, so it only handles internal grpc, and not external (like smm and account server calls)
- Implemented miiverse.v2 service from GRPC repo
- Account deletion does the following:
  - Delete all posts from the user
  - Delete all CDN content associated with those posts (paintings, screenshots)
  - Note: Account server CDN content is expected to be deleted by the account server
  - Note: Done in pages so we don't overload mongo with big deletions.
 
Almost fully tested, everything works. Was only not able to test screenshot deletions, but im just gonna hope that it works properly.


apps/miiverse-api/src/services/internal/server.ts
```diff
@@ -75,12 +78,12 @@ export async function setupGrpc(): Promise<void> {
...
 				payload: JSON.stringify(result.body)
 			};
-		},
-		sMMRequestPostId: async (_request) => {
-			throw new ServerError(Status.UNIMPLEMENTED, 'Not implemented');
 		}
 	});
 
+	// External communication (other pretendo services --> miiverse-api)
+	server.with(apiKeyMiddleware).add(MiiverseServiceDefinition, miiverseDefinition);
```
ashquarky, line 85: A nice improvement for the future would be to have a different (or 2nd-order function or whatever) API key check here, so that there can be unique keys for the various services.
Not really relevant right now but might be interesting if we ever want to open up the "internal" api for third-party clients

ashquarky (approved):
Haven't tested but code LGTM

PR #200, opened 2025-10-05
Feat: 3DS User Settings & Toolbar API

CaramelKat:
Resolves N/A

### Changes:

This introduces the user settings page on the 3DS. The larger part of this change includes a new api for configuring the toolbar on a page-by-page basis. The following data tags can be added to a given page, and the toolbar will be updated as appropriate:

| Tag | Action |
| --- | ------ |
|`data-toolbar-config`| This flag is used to identify the element with the rest of the config options|
|`data-toolbar-mode`| Int passed into the `cave.toolbar_setMode()` function |
|`data-toolbar-message`| String passed into the `cave.toolbar_setWideButtonMessage()` function |
|`data-toolbar-onclick`| Executes the function name passed into it when the wide button is clicked. Function name must be declared on the `window` object |
|`data-toolbar-onback`| Same as above, but for the back button |
|`data-toolbar-active-button`| Int passed into the `cave.toolbar_setActiveButton()` function |
|`data-toolbar-bgm`| Name of the BGM track to play when page is loaded |
|`data-toolbar-exit-bgm` | Name of the BGM track to play when the page is exited |
|`data-toolbar-sound` | Name of the sound effect to play when page is loaded |

I've tested this on my N3DS XL and things appear to be fine. If we want to tweak this api at all I'm open to suggestions. I attempted to make this fairly generic so it could be used elsewhere.

Also, since I'm not currently in a spot to overhaul all of the user pages to tsx I'm still using ejs for this. No new ones were added, so I think that's okay for now but I'd like to take a stab at the user pages here in the near future.

### DIFF
--- a/apps/juxtaposition-ui/src/webfiles/ctr/communities.ejs
+++ b/apps/juxtaposition-ui/src/webfiles/ctr/communities.ejs
@@ -3,7 +3,9 @@
 <%- include('partials/head', { title: lang.global.communities }); %>
 <body>
 <div id="body">
-    <header id="header" class="buttons">
+    <header id="header" class="buttons"
+		data-toolbar-config
+		data-toolbar-active-button="3"
         <h1 id="page-title" class=""><%= lang.global.communities %></h1>
         <a id="header-communities-button" class="right" href="/titles/all" data-pjax="#body"><%= lang.all_communities.text %></a>
     </header>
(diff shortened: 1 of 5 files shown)


Khentendo:
Will this include the option to change Game Experience from within the application on 3DS? Currently, we can only do this on the Wii U or during the First Time User setup on 3DS.

ashquarky:
Looks pretty solid, will do a proper review in the coming days

Issue #199, opened 2025-10-04
[Enhancement]: Fix Juxtaposition Embedding on Pretendo Forum


Joshua8600:
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What enhancement would you like to see?

Discussed in mod meeting; Juxtaposition posts do not embed on our Pretendo Network Forum

Quarky said it is likely the opengraphs that need fixing

Indeed, before submitting, the Forum says the oEmbed/OpenGraph tags that are missing are "description, image"

Issue #175, opened 2025-09-26
[Enhancement]: Allow users to delete their own posts on the web version


superfever3000:
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What enhancement would you like to see?

On the console versions, users can delete their own posts, but it's not possible on the web version since users can only see the report button on their posts.

Issue #148, opened 2025-07-22
[Bug]: Reply count on a post doesn't go down when deleting a reply


NewEmmettVA283:
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What happened?

The reply count on a post did not go down after a reply has been deleted. Any new replies are added to the total with the deleted replies.

### What did you expect to happen?

After deleting a reply, the reply count on the post would go down by one.

Issue #141, opened 2025-06-30
[Bug]: Deleting a post on web doesn't notify the target user


ashquarky:
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What happened?

.

### What did you expect to happen?

.

### Steps to reproduce?

.

### Other relevant information. (OPTIONAL)

Reported-by: @Joshua8600

Joshua8600:
This has not been fixed. Just tested.

Removing a post/comment from the page directly, rather from the reports page, still does not notify the user.

PR #202, opened 2025-10-11
feat(internal): Move post deletion and empathy to backend

### LINKED ISSUE #124: [Bug]: Putting an "&" in Post Removal (and Ignore) message deletes it and everything after it.
### Checked Existing

### What happened?

If a moderator deletes a post (or ignores a report) and the reason as an "&". it will delete it and everything after it.

Example:

If a removal/report reason is "Being mean & starting drama (Rule 1)." it will only show "Being mean "

This affects both community and moderators alike as we can't see full reasons for removing or ignoring a report.

For the whole message to be visible.

### Steps to reproduce?

1. Remove a post or ignore a report and put an "&" symbol in the reason with text after it.

_No response_

ashquarky:
Part of #112 
Resolves #124, #175, #148, #141 and maybe #199

### Changes:

Moves the post deletion to the miiverse-api gRPC backend, as well as fixing a boatload of bugs with post deletion.

Also moves the post empathy. This code is a little awkward since it needs to increment and decrement `empathy_count` in a non-racy way. If we were to replace the database `empathy_count` with taking the length of the `yeahs` array, this code could be dramatically simplified. I might do that!

Bonus random fix to the OpenGraph tags since apparently they didn't show up right on the forums.

### DIFF
--- a/apps/juxtaposition-ui/src/api/post.ts
+++ b/apps/juxtaposition-ui/src/api/post.ts
@@ -1,5 +1,4 @@
 import { apiFetchUser } from '@/fetch';
-import { deleteOptional } from '@/util';
 import type { ResultDto } from '@/api/result';
 import type { PageDto } from '@/api/page';
 import type { UserTokens } from '@/fetch';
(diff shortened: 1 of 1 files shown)


ashquarky:
I've kinda fallen out of love with the design of the middleware stuff, feel free to be harsh on it. The idea worked well for the internal API but is more clunky here

mrjvs:
So I completely forgot about this, mb. ill rereview it on sunday!

apps/juxtaposition-ui/src/middleware/checkBan.ts
```diff
@@ -0,0 +1,96 @@
...
+		if (request.guest_access || request.path === '/login') {
+			return next();
+		} else {
+			return response.status(401).send('Ban Check Failed: No user or guest access');
+		}
+	}
+
+	// Set access levels
+	response.locals.tester = account.pnid.accessLevel >= 1 && account.pnid.accessLevel <= 3;
+	response.locals.moderator = account.pnid.accessLevel == 2 || account.pnid.accessLevel == 3;
```
mrjvs: Can we use `accountIsModerator()` for this?

apps/juxtaposition-ui/src/middleware/webAuth.ts
```diff
@@ -0,0 +1,97 @@
...
+		if (pid !== null) {
+			pnid = await getUserAccountData(pid);
+		}
+
+		// Cache auth result
+		request.session.pnid = pnid;
+	} else {
+		// Use the cache
+		pnid = request.session.pnid;
+	}
```
mrjvs: This section could have less nesting, similar to what I suggested in `consoleAuth.ts`

apps/juxtaposition-ui/src/middleware/webAuth.ts
```diff
@@ -0,0 +1,97 @@
...
+		response.clearCookie('access_token', { domain: cookieDomain, path: '/' });
+		response.clearCookie('refresh_token', { domain: cookieDomain, path: '/' });
+		response.clearCookie('token_type', { domain: cookieDomain, path: '/' });
+	}
+
+	if (pnid) {
+		// populate account
+		response.locals.account = {
+			pnid,
+			settings: await getUserSettings(request.tokens),
```
mrjvs: Same comment as `consoleAuth.ts`, I feel like using the pid is safer.

PR #249, opened 2025-11-29
Implement metrics for user counts

mrjvs:
Changes:
- Moved metrics code into `metrics.ts` to centralize it
- Added various metrics regarding user counts
   - It's not possible to have one metric for all users, you can't remove duplicates if you only have one counter for one range
- Made admin routes that need metrics use the same functions as the metrics
- Added an index to settings for fast metric querying

Offtopic changes:
- added missing indexes that are on production but not yet in the code
- fixed linting for entire project
- Updated linting CI to not allow any warnings

### DIFF
--- a/apps/miiverse-api/src/database.ts
+++ b/apps/miiverse-api/src/database.ts
@@ -15,6 +15,7 @@ import type { HydratedPostDocument, IPostInput } from '@/types/mongoose/post';
 import type { HydratedCommunityDocument } from '@/types/mongoose/community';
 
 let connection: mongoose.Connection;
+mongoose.set('strictQuery', true);
 
 export async function connect(): Promise<void> {
 	connection = mongoose.connection;


apps/juxtaposition-ui/src/metrics.ts
```diff
@@ -0,0 +1,93 @@
...
+	}
+});
+
+export const activeYearlyGauge = new Gauge({
+	name: 'juxtaposition_active_users_yearly',
+	help: 'Users active in the last 365 days',
+	async collect(): Promise<void> {
+		const yearlyRangeMs = 365 * 24 * 60 * 60 * 1000;
+		const cutoff = new Date(Date.now() - yearlyRangeMs);
+		const count = await SETTINGS.countDocuments({ last_active: { $gt: cutoff } });
```
ashquarky: How often will Grafana poll this? This query seems kinda expensive and the yearly average should change very slowly..

ashquarky (approved):
Code works and is good

apps/juxtaposition-ui/src/metrics.ts
jonbarrow: My 2 cents, I agree with Quarky. Using an aggregate doesn't substantially change the code/complexity and gives us a free optimization, so I don't see a reason not to. The aggregate is arguably less clear than seeing `countDocuments` when reading things at a glance, but I'd expect anyone reading our metrics code should have a grasp on Mongo/Mongoose anyway (plus a little comment just noting that it's basically a faster `countDocuments` also solves that)

PR #255, opened 2025-12-09
Sort replies by oldest-first

mrjvs:
Changes:
- Fixes small bug introduced in #218 where replies were not ordered oldest-first

Master branch should now be deploy-able again, all bugs have been fixed.

Code was tested


ashquarky (approved):
Didn't test myself but code looks fine otherwise. I suppose we should add this to the other endpoints too going forward

PR #278, opened 2026-01-02
Fix deletion actions

### LINKED ISSUE #124: [Bug]: Putting an "&" in Post Removal (and Ignore) message deletes it and everything after it.

### LINKED ISSUE #141: [Bug]: Deleting a post on web doesn't notify the target user

mrjvs:
Changes:
- Fixed issue where deletion notifications did not trigger on dropdown delete (reason var was below the usage, triggered an error)
- Removed the deletion notifications from dropdown delete (joshua requested the behaviour stay the same after the fix)
- Renamed dropdown delete for moderators so it says "silently delete" instead of normal delete
- Fixed URL encoding problems with ignoring or resolving reports

Code is fully tested

Resolves #124 
Resolves #141

### DIFF
--- a/apps/juxtaposition-ui/src/webfiles/web/reports.ejs
+++ b/apps/juxtaposition-ui/src/webfiles/web/reports.ejs
@@ -1,6 +1,6 @@
 <html lang="en">
 <head>
-    <%- include('partials/head'); %>
+	<%- include('partials/head'); %>
 </head>
 <body>
 <div id="main">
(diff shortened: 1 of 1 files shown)


ashquarky (approved):
LGTM. it's always URLSearchParams

PR #279, opened 2026-01-04
fix: Use display:none to hide the toast so it won't eat mouseclicks

ashquarky:
at certain window sizes this overlaps with the login button, yikes!

Resolves #XXX

### Changes:


ashquarky (commented):
GitHub won't let me select "Approve" or "Request Changes" which is fine because I'm not really doing either of those things anyway.

It's OK to merge, I just think it could go a little further with the refactor

PR #284, opened 2026-01-07
Porting more to JSX and typescript

mrjvs:
Changes:
- Moved everything in routes/web to typescript
  - Ported login.ejs to JSX
  - It still uses error.ejs for error rendering
  - Also updated middleware to use JSX for login page rendering
- Ported redisCache.js to typescript
  - Fixed bug with expiration, library updated the signature for expiration setting
  - Fixed another bug where false was returned instead of null
- Ported juxt-web/routes/index.js and juxt-web/index.js to typescript
- Ported server.js to typescript
  - Removed `ttl` option from session(), the types indicate that isn't a thing with the current library version
  - Added types for cookie-parser lib
- Ported routes/console/show.js to typescript
  - Ported first_run.ejs to JSX
    - Moved the embedded CSS and JS to dedicated files
  - It still uses guest_notice.ejs, still needs porting later
  - Removed the check for the `1000000` pid. It wasn't possible to reach.
- Ported routes/console/web.js to typescript
- Ported endpoint mongoose model to typescript
- Ported report mongoose model to typescript
- Ported settings mongoose model to typescript
- Updated web.css so tabs are horizontally stacked instead of vertical

Notes:
- I tested all changes
- There seemed to be a lot of broken references to JS and CSS files in first_run.ejs - I wonder if it worked in the first place

Resolves a part of #12

### DIFF
--- a/apps/juxtaposition-ui/src/server.ts
+++ b/apps/juxtaposition-ui/src/server.ts
@@ -40,7 +40,6 @@ app.disable('x-powered-by');
 app.set('view engine', 'ejs');
 app.set('views', path.join(distFolder, '/webfiles'));
 app.set('trust proxy', config.http.trustProxy);
-app.get('/ip', (request, response) => response.send(request.ip));
 
 // Create router
 logger.info('Setting up Middleware');


apps/juxtaposition-ui/src/services/juxt-web/views/web/loginView.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+	return (
+		<WebLoginRoot head={extraHead}>
+			<div id="toast" data-show={props.toast ? 'true' : undefined}>
+				{props.toast}
+			</div>
+			<div className="wrapper">
+				<div className="account-form-wrapper">
+					<a className="logotype" href="https://pretendo.network">
+						<svg xmlns="http://www.w3.org/2000/svg" width="120" height="39.876">
```
ashquarky: Would be better to have the svg as its own TSX element (or even in its own file) to make it easier to swap out with a bundled version when the time is right for that

mrjvs:
> You picked all the weird ones with the custom head tags to do!

Yea, other than admin, this should all be the difficult stuff out of the way

PR #266, opened 2025-12-27
fix(ui): optimize images

suprstarrd:
Resolves #262. If I remember correctly, there's a few other related issues we can revisit here too.

### Changes:

This is a pull request that optimizes the images Juxt offers as best as possible.

Here's what @imgbot was able to do:
<details>
<summary>
Details
</summary>

| File | Before | After | Percent reduction |
|:--|:--|:--|:--|
| /apps/juxtaposition-ui/src/webfiles/web/images/add-post-no-image.png | 11.90kb | 4.00kb | 66.37% |
| /apps/juxtaposition-ui/src/webfiles/portal/images/add-post-no-image.png | 11.90kb | 4.00kb | 66.37% |
| /apps/juxtaposition-ui/src/webfiles/ctr/images/banner.png | 20.25kb | 9.69kb | 52.12% |
| /apps/juxtaposition-ui/src/webfiles/portal/images/banner.png | 20.25kb | 9.69kb | 52.12% |
| /apps/juxtaposition-ui/src/webfiles/web/images/banner.png | 20.25kb | 9.69kb | 52.12% |
| /apps/juxtaposition-ui/src/webfiles/portal/images/background.png | 233.20kb | 122.99kb | 47.26% |
| /apps/juxtaposition-ui/src/webfiles/ctr/images/bandwidthalert.png | 44.18kb | 28.74kb | 34.95% |
| /apps/juxtaposition-ui/src/webfiles/portal/images/bandwidthalert.png | 44.18kb | 28.74kb | 34.95% |
| /apps/juxtaposition-ui/src/webfiles/web/images/bandwidthalert.png | 44.18kb | 28.74kb | 34.95% |
| /apps/juxtaposition-ui/src/webfiles/portal/images/bandwidthlost.png | 111.69kb | 84.11kb | 24.70% |
| /apps/juxtaposition-ui/src/webfiles/web/images/bandwidthlost.png | 111.69kb | 84.11kb | 24.70% |
| /apps/juxtaposition-ui/src/webfiles/ctr/images/icons.png | 12.70kb | 10.94kb | 13.89% |
| /apps/juxtaposition-ui/src/webfiles/portal/images/splash-background.png | 43.35kb | 37.50kb | 13.51% |
| /apps/juxtaposition-ui/src/webfiles/web/images/splash-background.png | 43.35kb | 37.50kb | 13.51% |
[...]

PR #288, opened 2026-01-10
feat(ctr): Dynamically generate spritesheets

ashquarky:
Resolves #XXX

### Changes:

The old Juxt spritesheet system had a lot of up/down scaling of the sprites, background positioning mixed with normal styling, and other things that made it very hard to adjust any layouts involving sprites at all. This PR:

- Resamples each sprite according to the actual resolution it's rendered at on hardware
- Adds a script using the spritesmith library to create the sheet and suitable CSS on the fly
- Ports all pages using sprites to this new CSS stylesheet

Every effort was kept to keep the same layout (or slightly improve it) but like anything that touches Juxt's stylesheets, a bit of shifting around is kinda inevitable. Something something redesign something something new markup.

### DIFF
--- a/apps/juxtaposition-ui/nodemon.json
+++ b/apps/juxtaposition-ui/nodemon.json
@@ -4,7 +4,8 @@
   ],
   "ext": "js,ts,ejs,tsx,jsx,json,css",
   "ignore": [
-    "src/webfiles/ctr/css/sprites.css"
+    "src/webfiles/ctr/css/sprites.css",
+    "src/webfiles/ctr/images/sprites.png"
   ],
   "env": {
     "NODE_ENV": "development"


ashquarky:
My only concern with making it a build plugin is that anything we do with #266 needs to be automated *in the build process* rather than in the repo, since the spritesheet image itself is generated dynamically. The 3DS especially benefits from image optimisation.
Keeping it in the repo is more messy, but means we only have to re-run optimisations when something actually changes on the spritesheet.
Not sure if @lumiscosity and @suprstarrd have opinions here. oxipng is just, a command we can run, right? I could see about adding that to the build too...

ashquarky:
Oki. I can try to add a build step for this, at least for stuff that passes through ESBuild.

ashquarky:
If curious, here's the results of the new oxipng plugin against just the CTR assets:
```
SMTH: wrote 24 sprites
OXI tab.png: 124b -> 124b
OXI background.png: 437b -> 175b
OXI headline-green.png: 563b -> 556b
OXI header.png: 3592b -> 3592b
OXI headline.png: 581b -> 574b
OXI sprites.png: 14710b -> 6696b
```
Not bad!

mrjvs (changes requested):
I've tested everything and it looks good. Should be ready to merge after fixing the one tooling bug.

PR #295, opened 2026-01-18
Port `post.js` and `communities.js` routers to JSX + some more typescript ports

mrjvs:
Changes:
- Port `post.js` router to TS
  - I moved logic for `canPost` to the router instead of the view
  - That includes porting `post.ejs` to JSX
- Port `error.ejs` to JSX
- Port `error_fatal.ejs` to JSX
- Port `notifications.js` router to TS (not including JSX)
  - I changed the logic of marking notifications as read, it now does it before render instead of after
- Port `communities.js` router to TS
  - Includes porting `communities.ejs`, `all_communities.ejs`, `sub_communities.ejs` and `community.ejs` to JSX
  - caching works slightly differently, it now wipes the cache if it needs to recompute and theres a util that calculates stuff
  - The follow endpoint has been reworked. It was very janky logic
  - Moved `canPost` and `isUserFollowing` logic to the router instead of the template. So there is less data to send through
- Deleted old unused EJS files, I hope I didnt break anything

Notes:
- Code has been tested
- I do want to test the posts router a little more, but it ready for review at this stage
- After this, only 2 routers remain with EJS. And no more routers with posting logic :tada:

### DIFF
--- a/apps/juxtaposition-ui/src/middleware/checkBan.tsx
+++ b/apps/juxtaposition-ui/src/middleware/checkBan.tsx
@@ -45,11 +45,13 @@ export const checkBan: RequestHandler = async (request, response, next) => {
 
 	if (!accessAllowed) {
 		response.status(500);
-		if (request.directory === 'web') {
-			return response.jsx(<WebLoginView ctx={buildContext(response)} toast="No access. Must be tester or dev" redirect={request.originalUrl} />);
-		} else {
-			return response.jsx(<PortalFatalErrorView code={5989999} message="No access. Must be tester or dev" />);
-		}
+		const banMessage = 'No access. Must be tester or dev';
+		const banCode = 5989999;
+		return response.jsxForDirectory({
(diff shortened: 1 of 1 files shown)


apps/juxtaposition-ui/src/middleware/checkBan.tsx
```diff
@@ -46,10 +48,7 @@ export const checkBan: RequestHandler = async (request, response, next) => {
 		if (request.directory === 'web') {
 			return response.jsx(<WebLoginView ctx={buildContext(response)} toast="No access. Must be tester or dev" redirect={request.originalUrl} />);
 		} else {
-			return response.render('portal/error_fatal.ejs', {
-				code: 5989999,
-				message: 'No access. Must be tester or dev'
-			});
+			return response.jsx(<PortalFatalErrorView code={5989999} message="No access. Must be tester or dev" />);
```
ashquarky: Missed one sorry

ashquarky (changes requested):

PR #304, opened 2026-01-20
fix: cleanup endpoint data

mrjvs:
Changes:
- Cleanup non-user data from `downloadUserData.json` endpoint
- Named variables properly, they were the wrong way around

Changes have been tested


ashquarky (approved):
Haven't tested but looks straightforward

PR #306, opened 2026-01-22
fix(ui): Refactor and improve XHR for empathy and post deletion

ashquarky:
Resolves #XXX

### Changes:

Refactors how XHR is handled in the client JS. This was initially intended as a CTR fix for the Cloudflare XSS trip, but the error handling was improved enough to justify porting it to all platforms.

Thankfully, due to the bundler modernising CTR's syntax a lot, it's tolerable to just copy-paste the code between the three, with the only big issue being var/const.

- Finish setting up TypeScript for the client JS. This includes eslint rules and types for the browser globals.
- Pull XHR code out into a seperate TypeScript file and modernise it a little (`onload` works - it's old Chrome, not IE6!)
- Split empathy and delete API calls into their own functions, a bit reminiscent of the gRPC stuff on the server side. Also improve the error handling for JSON parse errors and others.
   - The delete API doesn't actually return a JSON object from the server - fake it for now, I'll change the server route to be more sensible later.
- Work around an issue on CTR where only GET and POST requests work. (This is probably related to Nintendo's domain allowlist on this platform)
   - This works with a HTTP header to override the method. Infra: does "smuggling" request methods have security implications, especially via Cloudflare? Can limit to CTR endpoints if it's a problem.
- Unify the post deletion experience across platforms, with the same texts and error messages.
- Fix a CSS bug where the toast on web appeared behind the navbar in mobile layout.

This touches the post templates to remove the `onclick` handlers, so #295 will need equivalent changes to the TSX versions.

The XHR and API files are near-identical between platforms, so a good candidate for a `@repo/client-common` library. Food for thought.

### DIFF
--- a/apps/juxtaposition-ui/src/services/juxt-web/views/ctr/postPageView.tsx
+++ b/apps/juxtaposition-ui/src/services/juxt-web/views/ctr/postPageView.tsx
@@ -31,7 +31,7 @@ export function CtrPostPageView(props: PostPageViewProps): ReactNode {
 						: null}
 					{post.pid === props.ctx.pid
 						? (
-								<a id="header-communities-button" className="delete header-button right" href="#" data-post={post.id} evt-click="deletePost(this)">Delete Post</a>
+								<a id="header-communities-button" className="delete header-button right" href="#" data-button-delete-post={post.id}>Delete Post</a>
 							)
 						: (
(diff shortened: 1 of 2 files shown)


mrjvs (approved):
(yes, I am approving my own commits now)

PR #339, opened 2026-02-11
fix(ui): Image scaling tweaks for sharper pixels

ashquarky:
Resolves #XXX

### Changes:

General improvements to image scaling and handling.
- Introduce "big" images - 2x nearest upscales for Wii U, since that WebKit can't do `image-rendering` and thus always uses blurry filtering. For most of our images - especially paintings - this is terrible. Instead, 2x them serverside (so paintings become 640px) and tweak the portal layout to display them 1:1 at that size.
  - Also helps a lot with lower-res 3DS screenshots.
  - Increases page weight, but sharp rendering of paintings is kinda essential for the Miiverse aesthetic, so..
- Introduce 48px and 96px icons (used in loads of places in the UI, but can't actually use them yet since none exist on CDN. Migration needed?)
- Bump JPEG quality for screenshots (again)
- Got the aspect ratio of the 3DS top screen wrong in #161, oops
- Tweak 3DS community layout to display icons at 64px (unscaled)
- Don't send debug.js to Wii Us

We've stacked up a few changes to image rendering, but the frontend still has to support all the legacy options. If these quality settings sit well with the users, we may want to consider doing a bulk migration of the historical post attachments and replacing them on the CDN. Then we can assume all the formats we want are there and don't need to worry about fallbacks.

Filed as a draft against a work branch for early review and to get the right diff, can change base to `dev` once #332 lands.

### DIFF
--- a/migrations/2-correct-screenshot-aspect.js
+++ b/migrations/2-correct-screenshot-aspect.js
@@ -1,0 +1,14 @@
+// Due to a typo in image handling, some posts are marked as 5:4 when the true aspect is 5:3.
+// ---
+// This script is intended to run in mongosh:
+// $ mongosh mongodb://localhost:27017/mydb 2-correct-screenshot-aspect.js
+// It takes roughly 1 minute on production database
+
+print(`Update post screenshots`);
+const result = db.posts.updateMany(
+	{ screenshot_aspect: "5:4" },
+	{ $set: { screenshot_aspect: "5:3" } }
+);
+
+print(`Finished updating screenshots!`);
+print(`Corrected ${result.modifiedCount}/${result.matchedCount} screenshots!`);


apps/miiverse-api/src/types/mongoose/post.ts
```diff
@@ -14,10 +14,13 @@ export interface IPost {
...
 
 	painting?: string; // base64, can be empty or undefined
+	painting_img?: string; // url fragment (leading /), can be empty or undefined
+	painting_big?: string; // url fragment (leading /), can be empty or undefined
 	screenshot?: string; // url fragment (leading /), can be empty or undefined
+	screenshot_big?: string; // url fragment (leading /), can be empty or undefined
 	screenshot_thumb?: string; // url fragment (leading /), can be empty or undefined
 	screenshot_length?: number;
-	screenshot_aspect?: string; // '4:3' '5:4' '16:9'
+	screenshot_aspect?: string; // '4:3' '5:3' '16:9'
```
mrjvs: 5:4 would still exist in the db, right? I'd rather it keep the comment with some explanation just for the old posts, will be easier for when we migrate.

Alternatively we can just write a migration to update the db. Cuz if I understand it right, it was never 5:4, just a typo?

ashquarky: None of the clients supported screenshot uploads other than 16:9 (the aspect ratio tracking is futureproofing for 3DS uploads) so the only way it could exist in the DB would be hitting the API directly.

Unfortunately it looks like some users *did* do this. ([Random example.](https://juxt.pretendo.network/posts/r8SzX2W3FSSC6g7B9yhHx)) So yeah, I guess a DB migration is appropriate.

mrjvs (approved):

PR #377, opened 2026-02-22
Use React contexts/hooks + i18next for localization

### LINKED ISSUE #366: [Enhancement]: Use `react-i18next` for translations
### Checked Existing

Currently we rawdog JSON files for translations. Using a translation library gives us some benefits:
1. Fallback to english with missing keys
2. Proper placeholders. Translation libraries have placeholder support that includes pluralization and other fancy things required for other languages.

It will remove the need to pass the language objects into the templates everytime.

`react-i18next` is the most popular translation library for react. It has some fancy features which will be very useful to us:
1. Component interpolation with `<Trans>`. It can insert components like `<strong>` inside of translation keys: `Hey <0>John Doe</0>!`
2. Autocomplete for keys. It reads the JSON files for types and extrapolates keys, so you can't use missing keys.

All of that makes me think react-i18next is the best option for us.

_No response_

### LINKED ISSUE #368: [Enhancement]: Use contexts and hooks in JSX templates
### Checked Existing

Currently, we have a single `ctx` prop that is being passed into every component. This is called prop-drilling, there are better methods.
Using react contexts, you can pass information to all child components. This is a better way to pass information globally.

The following hooks can be made when using contexts:
- `useUser()` - Get user info: lang, perms, pid, any related settings
- `useCdn()` - Utils for generating CDN urls
- `useRequest()` - Get request info: user agents, request ID, etc
- `useCache()` - Get caching utils: userMap, communityMap

Language files will be replaced with a translation library (see #366)

_No response_

mrjvs:
Changes:
- Replace `ctx` prop with global context: `RenderContext`
  - Global context is automatically added to every rendered JSX view. I figured it doesn't hurt to always have
- Moved all common JSX utils into it's own `common` platform folder (using the newly discussed folder structure)
- Created a bunch of hooks to use the various variables from the RenderContext
  - `useUser`: Settings/variables about the current user
  - `useCache`: Ways to read from the various caches
  - `useRequest`: Get state from the request, currently only the useragent and the language
  - `useUrl`: Replacement for the url utils we've had
- Ported localization to i18next
  - Moved locales folder to a place I think is more suitable
  - Load in every language in the locales folder
  - Injected i18next instance into all JSX views
  - Added a `<T k="" />` component, a shorthand for using `<Trans i18nkey="" />` (cuz its too long and looks weird)
  - Keys are typechecked based on the english locale file

Most of this refactoring is done with some fancy regex+replaces. Very sorry about the size of the PR 😿 

Resolves #366 
Resolves #368


ashquarky:
yeah i'm gonna need some time for this one

apps/juxtaposition-ui/src/services/juxt-web/views/common/components/T.tsx
```diff
@@ -0,0 +1,8 @@
+import { t } from 'i18next';
+import type { ReactNode } from 'react';
+import type { Trans } from 'react-i18next';
+
+export type TranslationKey = Parameters<typeof Trans>[0]['i18nKey'];
+export function T(props: { k: TranslationKey }): ReactNode {
```
ashquarky: Any reason for this to be a formal component? I feel like we could do:
```tsx
<span class="existing-thing">{ t("ui.existing_thing"); }</span>
```
Then attributes and DOM contents would be the same. A raw string is ReactNode, right?

apps/juxtaposition-ui/src/services/juxt-web/views/common/hooks/useUser.ts
```diff
@@ -0,0 +1,22 @@
...
+	moderator: boolean;
+	developer: boolean;
+};
+
+export type UseUserValue = {
+	pid: number; // TODO make this `null` when not logged in
+	perms: UserPermissions;
+};
+
+export function useUser(): UseUserValue {
```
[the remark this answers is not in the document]
ashquarky: Also ideally we'd have a full GetUserDataResponse instead of just the pid, we always have it from the middleware anyway

apps/juxtaposition-ui/src/i18n.ts
```diff
@@ -0,0 +1,54 @@
+import fs from 'fs/promises';
+import path from 'path';
+import i18next from 'i18next';
+import { initReactI18next } from 'react-i18next';
+import { langsFolder } from '@/util';
+import type en from '@/assets/locales/en.json';
+import type { ParamPack } from '@/types/common/param-pack';
+
+const entries = await fs.readdir(langsFolder, { withFileTypes: true });
```
ashquarky: Reading these at runtime is a little spicy, they used to be compiled in by the bundler

apps/juxtaposition-ui/src/services/juxt-web/views/common/hooks/useUrl.ts
```diff
@@ -0,0 +1,50 @@
...
+import type { CommunitySchema } from '@/models/communities';
+
+export type CtrHeader = { bannerUrl: string; imageId: string; legacy: boolean };
+type Serializable = string | number | boolean | undefined | null;
+
+function buildUrl(u: string, query?: Record<string, Serializable>): string {
+	const [pathName, search = ''] = u.split('?', 2);
+	const originalParams = new URLSearchParams(search);
+	Object.entries(query ?? {}).forEach((entry) => {
+		if (entry[1]) {
```
ashquarky: Should do explicit undefined/null check here to not accidentally eat e.g. `include_replies=false` or `offset=0`

ashquarky (changes requested):
Just some design and codestyle nits. I'm assuming you've tested it (Cemu lets you change the console lang easily enough)

apps/juxtaposition-ui/src/services/juxt-web/views/common/hooks/useUrl.ts
```diff
@@ -0,0 +1,50 @@
...
+	// TODO add util for building a mii image
+};
+
+export function useUrl(): UseUrlValue {
+	const ctx = useRenderContext();
+
+	return {
+		ctrHeader: com => getCtrHeader(ctx.cdnUrl, com),
+		url: buildUrl,
+		cdn: path => buildCdnUrl(ctx.cdnUrl, path)
```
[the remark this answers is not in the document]
mrjvs: Yea, we can add a bunch of url building utils. Perfect place for it

apps/juxtaposition-ui/src/util.ts
```diff
@@ -531,6 +507,7 @@ export function humanFromNow(date?: Date | DateTime | string | null): string {
 const filename = fileURLToPath(import.meta.url);
 // The root of the dist/ folder.
 export const distFolder = path.dirname(filename);
+export const langsFolder = path.join(path.dirname(filename), 'assets/locales');
```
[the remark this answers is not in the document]
mrjvs: Fixed

apps/juxtaposition-ui/src/services/juxt-web/views/common/hooks/useUrl.ts
mrjvs: Fixed

ashquarky (dismissed):

ashquarky (approved):
LGTM! Mammoth effort

PR #378, opened 2026-02-23
feat(admin): Mod panel tweaks + improved title ID input control

ashquarky:
Resolves #XXX

### Changes:

Future Changes are going to require all communities to have accurate TitleIDs set, so here's some changes for that and general mod panel tweaks.
- Change Icon and MiiIcon to the new component style, with folder structure and a `type` parameter for variants. Inline SVG icons are now called "UIIcon" to avoid the name clash.
- Add extra details to report rendering as requested by mods (PID view, link to reporter page, timestamp details)
- Bundle admin.js with main js so it's one large script. Only affects admin view.
- Port mod panel client code to TypeScript and refactor all event handlers into the new style with `data-` attributes.
- Make search bars into a component and use normal HTML forms instead of JS.
- Improve TitleID control so it can accept decimal, hex and dashed format TitleIDs. Hex and dashed IDs are commonly found on community wikis. IDs will always be displayed in dashed format.

Dashed, hex and decimal formats: 
image

All displayed as hex after saving: 
image

- [x] I have read and agreed to the [Code of Conduct](https://github.com/PretendoNetwork/Pretendo/blob/master/.github/CODE_OF_CONDUCT.md).
- [x] I have read and complied with the [contributing guidelines](https://github.com/PretendoNetwork/Pretendo/blob/master/.github/CONTRIBUTING.md).
- [ ] What I'm implementing was an [approved issue](../issues?q=is%3Aopen+is%3Aissue+label%3Aapproved).
- [x] I have tested all of my changes.

PR #403, opened 2026-03-05
feat(ui): Restrict shot upload according to title id and community

ashquarky:
Resolves #XXX

### Changes:

- Only allow posting screenshots in their designated communities. This is checked based on the community's title_id and the suspended software's title ID (from the param pack).
- Extra title IDs can be provided in a new shot_extra_title_ids field. This is for games that don't have communities yet but we want to allow screenshots in a related community; without polluting the API response for title IDs.
- A new shot mode toggle is added. This allows entirely blocking screenshots in a community (e.g. 3DS games that use the cameras) or forcing screenshots to be enabled (e.g. MK7 or PPT that disable screenshots for no reason).

This is ready for review but DO NOT merge until #378 has been in prod for a week or so - mods will need time to audit and fix the title IDs on our existing communities (plenty of them are wrong).

This is the final major feature needed for screenshot upload on CTR to be submitted.

### DIFF
--- a/apps/juxtaposition-ui/src/services/juxt-web/routes/admin/admin.tsx
+++ b/apps/juxtaposition-ui/src/services/juxt-web/routes/admin/admin.tsx
@@ -6,7 +6,7 @@ import { deletePostById, getPostById } from '@/api/post';
 import { database } from '@/database';
 import { uploadHeaders, uploadIcons } from '@/images';
 import { logger } from '@/logger';
-import { COMMUNITY } from '@/models/communities';
+import { COMMUNITY, CommunityShotModes } from '@/models/communities';
 import { POST } from '@/models/post';
 import { SETTINGS } from '@/models/settings';
 import { humanDate, createLogEntry, getReasonMap, getUserAccountData, newNotification, updateCommunityHash } from '@/util';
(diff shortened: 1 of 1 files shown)


apps/juxtaposition-ui/src/models/communities.js
```diff
@@ -112,6 +112,14 @@ export const CommunitySchema = new Schema({
 	permissions: {
 		type: PermissionsSchema,
 		default: {}
+	},
+	shot_mode: {
+		type: String,
+		default: 'allow'
+	},
```
mrjvs: We should probably make this an enum in mongoose, that way we can be sure that the values are what we expect them to be.

Enum can be reused by the zod schemas

apps/juxtaposition-ui/src/services/juxt-web/views/web/newPostView.tsx
```diff
@@ -50,6 +50,7 @@ export type NewPostViewProps = {
 	url: string;
 	show: string;
 	messagePid?: number;
+	shotMode: string; // 'allow' | 'block' | 'force'
```
mrjvs: We can use the actual type here

PR #424, opened 2026-03-10
Frontend modules for CTR

mrjvs:
Changes:
- Introduced frontend modules - a simple module system to segment frontend logic into neat modules (for all platforms)
  - This system can hydrate partial documents, so this will make it harder to make memory leaking mistakes
  - The system includes a warning popup for hydration (since performance takes a hit if double hydration is ever done)
- Ported all CTR frontend code into modules
  - Where possible, the frontend code was refactored to use less anonymous functions
- Ported `juxt.js` to typescript
- Updated eslint config:
  - Now allows path aliases in webfiles
  - Now prevents using unsupported syntax for consoles (let/const/`...`)
  - Reorganized it a little

Note:
- Still needs testing


apps/juxtaposition-ui/webfiles/ctr/js/modules/more-posts.ts
```diff
@@ -0,0 +1,26 @@
+import { createModule } from '@repo/frontend-common';
+import { GET } from '@/js/xhr';
+import { modules } from '@/js/module-container';
+
+export var morePostsModule = createModule({
+	id: 'more-posts',
+	selector: '.load-more[data-href]',
+	hydrate: (ctx) => {
+		ctx.el.addEventListener('click', function (_e: Event) {
```
ashquarky: anonymous function

ashquarky (commented):
Overall quite like the design, just because it's frontend though I have some notes about extra flexibility you've added (deep arrays, optional hydration) that I don't think we actually need/use. Since bundle size is so important I always like to limit features/flex to what we actually use, knowing we can add stuff later if needs change.

apps/juxtaposition-ui/webfiles/ctr/js/modules/more-posts.ts
ashquarky: The potential for double-added handlers still feels like a big footgun that we could refactor our way out of with some code style requirements. I would like to avoid it unless it's really hard to factor out (using e.currentTarget)

ashquarky (dismissed):
I like the design! Should save this for immediately after the currently-queued release I think

mrjvs:
Superseeded by #605 which makes use of the new common library introduced in #604

PR #446, opened 2026-03-15
Express 5

### LINKED ISSUE #123: Upgrade to express 5
Express 5 has been released and it has a bunch of fixes and a few new features.

A migration guide has been written by the authors of express: https://expressjs.com/en/guide/migrating-5.html

### LINKED ISSUE #236: [Enhancement]: Node 24
### Checked Existing

Juxt runs on NodeJS 22, which ended active support in October and is now only receiving bugfixes. See https://endoflife.date/nodejs

We are still getting bugfixes and security patches, but at some point we should consider moving to the current active release (Node 24). 

I think this would include:
- Bumping the relevant Dockerfiles
- Updating `@types/node` (and resetting Dependabot ignore rules on this dep)

Security patches for 22 continue until 30 Apr 2027, so we can take our time here if we'd prefer, but I suspect we'll start running into libraries requiring the newer features if we take too long.

mrjvs:
Changes:
- Update to Node 24
- Update to express 5 using a [migration guide](https://expressjs.com/en/guide/migrating-5.html).
  - Note: the `app.listen()` error change seems to be a lie, I cannot reproduce it and typescript says it doesnt exist
- Migrate away from `express-subdomain` and use the same code and technique as website and account repo uses
  - Breaking change: juxt-dev domains dont work by default anyway, they need a config change
- Remove dependabot.yml. It's too spammy, now it will only make PRs for security vulnerabilities. We will do manual updates for other libraries.

I've ran both servers and clicked through a bunch of juxt pages, seems to work fine

Resolves #123 
Resolves #236


ashquarky:
I propose we review this after #450

PR #462, opened 2026-03-24
feat: add profile visibility

### LINKED ISSUE #394: [Feature]: Ability for a User to Private Their Profile to Guest Mode
### Checked Existing

The ability for users to "private" their profile so people viewing Juxt in Guest Mode (not being signed in and just viewing the profile page from the direct URL) cannot see their posts or any other information.

We recently received a User Report where an off-platform conflict led to their Juxtaposition profile being viewed by someone who is banned. The ability to Private their profile to Guest Mode would not only let this user feel safer on the platform, but it is a generally good privacy feature as a whole.

_No response_

suprstarrd:
Resolves #394

### Changes:

Allows users to make a profile private. Draft PR so doesn't work quite right yet.


suprstarrd:
Noting this as still of interest

ashquarky:
image

- Adds tickboxes to the account settings pages to set a new profile_visibility database field. This has the possible values "public" and "users_only" to allow extensions later.
- Introduce a "tryGetUserData" function to the Internal API. This tries to get the AccountData for an active user; considering the permissions of the requester. It bails out if the user is not found, deleted, banned, or private (and requested by a guest) and is a good spot to add user->user blocking later. 
- Use the new tryGetUserData function anywhere one user's data is exposed to a guest or a third-party user - except for feeds, as these aren't really viable without joins - such that they error out or return empty results for users the requester shouldn't be able to see. Most of the endpoints are defense-in-depth since the UI won't call them if the initial profile request fails, but adding it anyway in case other people start hitting this API or we restructure the frontend later.
- Adds a smaller version of the Bandwidth lost icon for web. It's displayed at 64x64 usually, but 128x128 for Retina and because that's the size of the account server Mii images.

apps/juxtaposition-ui/src/services/juxt-web/routes/console/userpage.tsx
mrjvs, line 126: I know you didnt change this, bit let's await `updateComment` to prevent a process crash

apps/miiverse-api/src/models/settings.ts
```diff
@@ -53,6 +53,10 @@ const SettingsSchema = new Schema<ISettings, SettingsModel, ISettingsMethods>({
 		type: Boolean,
 		default: true
 	},
+	profile_visibility: {
+		type: String,
+		default: 'public'
+	},
```
mrjvs, line 60: Let's add an enum constraint here so we know what values we can expect from the db

mrjvs (changes requested):
First half the review, still haven't looked at everything

apps/miiverse-api/src/services/internal/utils/user.ts
```diff
@@ -0,0 +1,38 @@
+import { Content } from '@/models/content';
+import { Settings } from '@/models/settings';
+import { errors } from '@/services/internal/errors';
+import { getUserAccountData } from '@/util';
+import { accountIsDeveloper, accountIsModerator } from '@/services/internal/middleware/auth-populate';
+import type { AccountData, ActiveAccountData } from '@/types/common/account-data';
+
+export async function tryGetUserData(auth: AccountData | null, targetPid: number): Promise<ActiveAccountData> {
```
mrjvs: I'm not sure about this method:
- From the name `tryGetUserData` it isn't obvious to me that it's doing user-preference based filtering
- It's fetching a lot of data, also from third parties - most of the data is not needed where it's called.

If the main goal was to unify the "can this user access this other user" logic, I would rather have a specific method that checks it: `canAccessUser(auth, targetSettings)`. (This does have the downside of needing to explicitly check it everywhere, but I still think that it's better to have it done explicitly instead of hidden)

As for the PNID data, I think it's best we solve that through smart DB use in the future. Then we can also apply it to feeds

mrjvs (changes requested):

apps/miiverse-api/src/services/internal/utils/user.ts
ashquarky: I'll have to stew on how to design it (it can't just return bool because we need to propogate the specific error code on some paths.. but not others..) but I think this is okay yeah.
Leaving PNID as a temporary fix? Since I assume that kind of query would be post-Postgres..

ashquarky:
This conflicts with #517 now so probably best to review one or the other, and I'll fix the conflict in the remaining one after it lands. Then can review it as it's ready to merge.

Sorry for stacking so many features up ^^;

mrjvs (dismissed):
Looks good now, nice work!

PR #466, opened 2026-03-25
Add Extra Checks for Duplicate Posts.

### LINKED ISSUE #99: [Enhancement]: Remove, or at least improve, "duplicate post" checks

imora7024:
Resolves #99

### Changes: Add Extra Community Checks for In-game and UI posts.

Okay. Im not exactly a coder to say much for this, i also don't have a local hosted environment to actually test this.. So i might need some help for now.

This Pull request might potentailly fix the issues caused by drawings specifically for In-game posts or UI related posts being duplicated and locking you permanently. This just adds 1 simple check for both ends. the Community ID and Olive ID.
For whatever reason. the current way we do this dosen't implement this and seems to cause issues with users (including me) Getting locked out of paintings.

For right now i don't have a way to test this at all. So please let me know if it needs any changes!

### DIFF
--- a/apps/miiverse-api/src/database.ts
+++ b/apps/miiverse-api/src/database.ts
@@ -102,7 +102,7 @@ export async function getDuplicatePosts(pid: number, post: IPostInput): Promise<
 	verifyConnected();
 
 	return Post.findOne({
-		community_id: community.olive_community_id,
+		community_id: post.community_id,
 		pid: pid,
 		body: post.body,
 		painting: post.painting,


imora7024:
Okay. so the Community ID is specifically used for Sub Communities. Any other checks that are needed for now?

PR #479, opened 2026-04-12
[part 1] Add community endpoints + revise internal API structure

mrjvs:
Changes:
- miiverse-api internal API:
  - Created a typed variant of an express router:
    - It enforces zod schemas that get passed in (body, query, params, response)
    - Forces the endpoint to have a guard (authorization middleware)
    - Gives a better request context (query, body, params, auth) to the handler
    - Handler returned data becomes the response
  - Routes using the new router now generate an openapi specification on build time
    - Make sure you give your zod schemas a type name using `.openapi('TypeNameHere')`
  - Ported existing internal API endpoints to the new router
  - Added community endpoints to internal API
  - Removed postman unit tests
  - Removed test router, it did nothing
  - Removed all old packages
  - Moved all global configuration into setup functions (so that the new build step doesn't need a full on working config to compile)
- Juxtaposition-ui changes:
  - Replaced `apiFetch` with a pure fetch-compatible version in `src/api/client`
  - Replaced everything in the `src/api` folder with a generated API client. Generation is done automatically on build
    - This does mean that miiverse-api will have to be compiled at least once before juxtaposition-ui is ran
  - Added the new API client to the request: `req.api`. It is preauthenticated with the current user.
  - Replaced all calls to the internal API with the new API client
  - Replaced all calls to community DB model with the new internal API endpoints
  - Replaced all usages of `InferSchemaType<typeof Community>` with internal API type
  - Replaced all usages of `PostDto` with internal API type

Testing checklist:
- [x] API client generation
- [x] Existing post endpoints
- [x] Frontend usage of existing post endpoints
- [x] Authentication of new API structure
- [x] New community endpoints
- [x] Frontend usage of new community endpoints
- [x] Does grpc + s3 still work? (check getUserData specifically)

Breaking changes:
   - There is now consistent sorting on community listing pages
   - Subcommunities listings are now limited to just type `1` (subcommunity)
   - Subcommunity page is limited to 90 items and not infinite

Resolves a little bit of #112 - the structure of this API will make it dead simple to add new endpoints though


apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.tsx
```diff
@@ -156,24 +152,24 @@ postsRouter.get('/:post_id', async function (req, res) {
...
 		}
 		return res.redirect(`/posts/${parent.id}`);
 	}
-	const community = await database.getCommunityByID(post.community_id);
+	const { data: community } = await req.api.communities.get({ id: post.community_id });
 	if (!community) {
 		return res.redirect('/404');
 	}
 
 	// increase limit for post replies since there's no pagination yet
```
ashquarky, line 171: Comment is now pointless

apps/juxtaposition-ui/src/api/client.ts
```diff
@@ -0,0 +1,72 @@
...
+		headers: JSON.stringify(Object.fromEntries(req.headers.entries())),
+		payload: req.body ? await new Response(req.body).text() : undefined
+	}, {
+		metadata
+	});
+
+	// Mask 404's as a succesfull `null` response
+	if (grpcResponse.status === 404 && req.method === 'GET') {
+		grpcResponse.status = 200;
+		grpcResponse.payload = JSON.stringify(null);
```
ashquarky, line 38: Kinda weird. Is there a callsite where we expect this?

apps/miiverse-api/src/services/internal/routes/communities.ts
```diff
@@ -0,0 +1,165 @@
...
+
+communitiesRouter.get({
+	path: '/communities/popular',
+	name: 'communities.listPopular',
+	guard: guards.guest,
+	schema: {
+		query: z.object({
+			limit: z.coerce.number().max(32).default(9)
+		}),
+		response: z.array(communitySchema)
```
ashquarky: pageDto?

apps/miiverse-api/src/services/internal/routes/communities.ts
```diff
@@ -0,0 +1,165 @@
...
+			olive_community_id: communityId,
+			title_id: titleId,
+			...typesToFilter
+		}));
+		if (!community) {
+			throw new errors.notFound();
+		}
+
+		// Unsure why this exists, possible legacy data?
+		if (!community.permissions) {
```
ashquarky: `{ permissions: {$not: { $type: 3 } } }` gives 0 results, so we can drop this check

ashquarky (changes requested):
Really good overall

apps/juxtaposition-ui/src/api/client.ts
mrjvs: For every GET endpoint that can return 404 I want it to return null instead of having to handle the error per API call. Unfortunately this seems to be the only point where you can customize the behaviour. It's a little janky but I can't find a better way to handle it.

apps/miiverse-api/src/services/internal/routes/communities.ts
mrjvs: I can't really add pagination to the endpoint due to the way it's calculated. and I dont think its right to return with page if there's no pagination.

It's definitely something to improve, but it probably needs a rework of the database for that to be possible.

apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.tsx
mrjvs: Oops, will add the `500` limit back

apps/miiverse-api/src/services/internal/routes/communities.ts
mrjvs: Fixed

apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.tsx
mrjvs: Oh I misread, I did have the 500 limit in the code.
The comment is still relevant since the frontend does no pagination. The backend always had it.

apps/miiverse-api/src/services/internal/routes/communities.ts
ashquarky: I guess it's just odd that we sometimes have [] and sometimes have .items, y'know? It's not consistent on the output side

apps/juxtaposition-ui/src/api/client.ts
ashquarky: I think it's an odd design choice to eat 404s but I'll let you cook on it

ashquarky (dismissed):

apps/miiverse-api/src/services/internal/routes/communities.ts
mrjvs: Introduced `listDto` for returning plain lists.

I still want to keep it separate from pageDtos, since they usually come with some extra stuff (totals, current page, cursors, etc). I introduce some extra stuff in the other PR's

ashquarky (approved):

PR #480, opened 2026-04-12
[part 2] Port activityfeeds datasource to miiverse-api

mrjvs:
Changes:
- Port fresh/popular/people feeds to miiverse-api
- Separate pagination into feeds (no total, can be cursor based in the future) and normal pagination (total, offset based)
- Removed unused database.js methods

Notes:
- **Breaking change**: Feeds now contain removed posts (if user is moderator), I've checked with moderators, they say its fine
- Post listings do not check if you have access to the community (private communities), this was already the case so I didn't change it.

Changes have been tested locally.

Resolves a part of #112


ashquarky (approved):
LGTM

PR #481, opened 2026-04-12
[part 3] Port follower and current user datasource to miiverse-api

mrjvs:
Changes:
- Added follower/unfollow endpoints, also implemented into frontend codebase
- Added follower/following listing endpoints, also implemented into frontend codebase
- Added endpoint to list a users posts, also implemented into frontend codebase
- Added endpoint to get the current user:
  - It does not error when banned, since it's used to get current account status as well
  - It's requested every frontend page and is the current auth state of the request
  - It does a few miscellaneous tasks: update mii name, set last_active, clear ban if date has passed
  - It's the replacement for getting the current user settings and current user content
- Updated the guest guard, it now ignores bans as well. The user guard now checks it

Notes:
- Depends on #480 - dont merge until that PR is merged

Resolves a part of #112


ashquarky (dismissed):
LGTM

mrjvs:
Gonna wait for #494 before merging, otherwise we have another suite of tests to run before it

PR #483, opened 2026-04-18
Remove DMs from frontend

mrjvs:
Changes:
- Remove database calls and routes related to messages from juxtaposition-ui
- Remove views related to messages
- Remove message count from notifications.json
- Update navigation so messages aren't there anymore
  - Web: remove messages nav item
  - Portal: Replace messages with friend requests in navigation. Remove friend requests from notifications page
  - Ctr: Remove tabs from news page
- Remove text related to messages from en.json, weblate will remove the keys from other files automatically
- Added migration to remove messages
- Removed JS related to updating the message-count-badge in the navigation

Notes:
- The miiverse-api project stays as-is, it still supports messages. Users just can't view it in the app anymore.


ashquarky (approved):
LGTM. Does the Friend Requests feature actually.. Work good..? I've never looked at it. We could consider putting one of the feeds or account settings there too if it's a bit meh.

mrjvs:
> Does the Friend Requests feature actually.. Work good..?

No idea tbh, We can see how well it works in prod and decide what to do with the navitem based on that

PR #492, opened 2026-04-26
Add basic automod functionality

### LINKED ISSUE #20: [Feature]: Auto-mode style moderation to assist human moderators
### Checked Existing

Add an auto-mod style system which will check UGC, and flag content for review automatically. Human moderators can then access flagged content quicker, leading to faster resolutions.

Moderation has been a hot topic recently, especially on Discord, and with our growing number of users it seems paramount that we try and make the mods lives a bit easier where we can. There is also the fact that Nintendo has been confirmed to be watching us and has spoken out with regards to our moderation, so we need to take it more seriously.

Right now we have to rely on human reports, or mods stumbling on posts, in order for action to be taken. While this works in small numbers, we are growing rapidly and it's time to start exploring our options on how to make this smoother/faster. We already use Discords automod feature, and Chubby has NSFW checks for images/links, for moderating the Discord, so adding this to Juxtaposition also makes sense.

An easy quick-and-dirty (partial) solution to this is a word/phrase/regex blacklist. This is how automod on Discord works (iirc). This would get us part of the way there, but not fully.

The next solution would be to try and use some sort of language processing. [NLP](https://en.wikipedia.org/wiki/Natural_language_processing) is a technique which has been around for decades, and might be useful. Though it is rather limited in what it can do. It's a complex topic, but for our purposes it would boil down to "train on input data, get classifications" (this is different than modern LLMs, mind you, which need MUCH MORE training data, are a black box, and are mostly built for chat-like interactions). This would get us likely better results than just a blacklist, but takes more to setup and is still limited on what it can do.

After basic NLPs/traditional classifiers, the next solution would be some form of advanced LLM. Let me make this ***STARKLY, EXTREMELY, CLEAR***, I am, without question, against the modern AI craze. I do not subscribe to the idea that modern AI tools/LLMs will/should replace humans, and I believe modern tools like Claude and ChatGPT to be harmful not just to the planet but to the industry/people as a whole and they leech off the creativity and work of real humans without any sort of compensation. That being said, it is something that exists now and is an option to be put here. I do not believe we should rely on harmful services like ChatGPT, but running a local instance of a model built from public datasets IS something we could consider? I have been told by Quarky that Llama runs decently well these days locally, but I still have SOME concerns over things like speed and scalability (it might work well when you're running a simple chat example, where data can come back whenever it needs to, but that doesn't work in real-time environments where we may need to scale). However these also have issues to consider, such as the fact that they weren't made with classification in mind so they must be prompted very specifically and runs the risk of people escaping the prompt and such. Products like https://sightengine.com come to mind, as they provide a VERY FAST, VERY ACCURATE, suite of APIs for content moderation such as detection porn/gore in images, PII/hate speech in text, etc. (their entire business model is essentially "moderation as a service"). That being said, there's MAJOR issues with SightEngine which make it unusable by us (I only bring them up as an example of what we should strive for in our own API/tools):

1. It's not open source, so we have no idea how the model works or is trained and we can't run it locally. This ties our moderation entirely to the whims of this company, and we have no idea if their model training was ethical or not.
2. ToS has broad data collection clauses. Even if they are just "cover our asses" type of clauses, it still doesn't look good coming from an AI company.
3. They seem to be pro "remove human moderators", which we disagree with. One of their selling points is "Reduce your costs by performing your moderation tasks with an automated solution" which is scummy.
4. It's incredibly expensive at scale. Free plan is only 2k operations per month with a max of 500 per day. Lowest paid plan is $30/m for 10k requests per month then $0.002 per request after. Next is $99/m for 40k requests per month then $0.002 per request after. Next is $399/m for 200k requests per month then $0.0015 per request after. And then "contact us" billing, which is always $$$$$$.
5. It would probably be terrible optics from a user's perspective. We don't want users thinking "Pretendo is sending our data to some rando AI company!!!", and as an end-user myself I would feel that exact way if I found out a service was doing this. This is one of the biggest killers imo.

That being said, we CAN likely get decently far on our own should we want to. There ARE places to get open datasets, such as https://www.kaggle.com/datasets which provides datasets for many different applications each with their own licenses (from what I understand, this is all ethically sourced and okay to use?). We could take some training sets there, train our own models for things like PII/hate speech/whatever detection, and then run the models ourselves (assuming we have hardware capable of doing so at a sufficient speed).

The thing we would lose though is image checks. But we might still be able to work around that as well. Chubby has a NSFW detection feature, and while it still has issues at times it's been working quite well.

There's also [PhotoDNA](https://www.microsoft.com/en-us/photodna), which is a product provided by Microsoft to help identify harmful images like CSAM. We should definitely consider applying for this at some point, regardless of what is decided here.

The other issue we run into with training our own models is that if we want to detect something that *doesn't* have a dataset, we need to build our own. And building a large enough dataset on our own is no easy task. For example, Kaggle has 0 datasets relating to software piracy. That means if we wanted to detect people talking about piracy we would need to create our own dataset to train the model with. One option would be to use data from our users, such as on Discord by using the results from automod, but they might see that as unethical. Another solution may be to generate fake data, such as asking another LLM (again, within ethics) to generate X number of fake messages based on the given context, but that might have the issue of "AI training on AI" resulting in worse results.

mrjvs:
Changes:
- Add automod rules and logs to db
  - two modes: log and block
  - one type: keyword (match by keywords)
  - can be disabled if needed
  - Has a title, and optionally a description for some extra context
- Add automod evaluation to posting flows
- Added a way to give user feedback when posting goes wrong, implemented for automod
- Added mod pages to view logs and rules (for moderators)
- Added mod pages to create/update/delete automod rules (for admins)

The html is incredibly janky and stupid, but I really cba to do CSS rn.

Resolves #20

### DIFF
--- a/apps/juxtaposition-ui/src/assets/locales/en.json
+++ b/apps/juxtaposition-ui/src/assets/locales/en.json
@@ -99,6 +99,7 @@
     "none": "No Friend Requests"
   },
   "new_post": {
+    "automod_error": "Your post contains text that is not allowed on Juxtaposition. For more information, please visit https://preten.do/juxt-rules",
     "new_post_short": "Post",
     "post_to": "Post to {{user}}",
     "swearing": "Post cannot contain explicit language.",
--- a/apps/juxtaposition-ui/src/models/automodLog.ts
+++ b/apps/juxtaposition-ui/src/models/automodLog.ts
@@ -1,0 +1,41 @@
+import { Schema, model } from 'mongoose';
+import type { HydratedDocument } from 'mongoose';
+
+export const automodAction = ['blocked', 'logged'] as const;
+export type AutomodAction = (typeof automodAction)[number];
+
+export type AutomodLog = {
+	rule_id: string;
+	created_at: Date;
+	action: AutomodAction;
+	post_id: string | null;
+	post_content_body: string | null;
+} & Document;
+
+export type HydratedAutomodLogDocument = HydratedDocument<AutomodLog>;
+
+export const automodLogSchema = new Schema<AutomodLog>({
+	rule_id: {
+		type: String,
+		required: true
+	},
+	created_at: {
+		type: Date,
+		required: true
+	},
+	action: {
+		type: String,
+		enum: automodAction,
+		required: true
+	},
+	post_id: {
+		type: String,
+		required: false
+	},
+	post_content_body: {
+		type: String,
+		required: false
+	}
+});
+
+export const AutomodLog = model('AutomodLog', automodLogSchema);
--- a/apps/juxtaposition-ui/src/models/automodLog.ts
+++ b/apps/juxtaposition-ui/src/models/automodLog.ts
@@ -7,6 +7,7 @@ export type AutomodAction = (typeof automodAction)[number];
 export type AutomodLog = {
 	rule_id: string;
 	created_at: Date;
+	author: number;
 	action: AutomodAction;
 	post_id: string | null;
 	post_content_body: string | null;
@@ -26,6 +27,10 @@ export const automodLogSchema = new Schema<AutomodLog>({
 	action: {
 		type: String,
 		enum: automodAction,
+		required: true
+	},
+	author: {
+		type: Number,
 		required: true
 	},
 	post_id: {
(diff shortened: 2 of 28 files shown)


ashquarky (approved):
Seems fine

PR #493, opened 2026-04-27
fix(ui): santise topic tags in URLs

### LINKED ISSUE #282: [Bug]: Topic tags are not sanitised inside a URL
### Checked Existing

### What happened?

When opening clicking on topic tag on WiiU (may also be on 3DS) that has a special symbol (like a `+`). It doesn't show any posts, because everything after the `+` is interpreted as another query param.

Clicking on a topic tag should show all of its posts regardless of special characters

### Steps to reproduce?

1. Click on topic tag of a post with a special character in the topic tag
2. Witness that there are no posts, even though there should be

This can be solved by doing `encodeURIComponent()` or using `URLSearchParams` when building the URL. There will be multiple places where this is a problem.

ashquarky:
Fixes #282


apps/juxtaposition-ui/src/services/juxt-web/routes/console/topics.tsx
mrjvs: I think this also needs it

mrjvs (changes requested):

mrjvs (approved):

PR #482, opened 2026-04-15
[part 4] Feat/port admin datasource

mrjvs:
Changes:
- added endpoints to list/create/view/edit communities (for admins pages), also implemented into frontend
- added endpoints to manage reports (list/ignore/removePost), also implemented into frontend
  - Breaking change: now only lists 150 reports maximum
- moved post deletion audit logs to miiverse-api
- added endpoints to list users (for admin pages), also implemented into frontend
- added endpoints to view user profiles, also implemented into frontend
  - Breaking change: moderators can no longer see their data when user is banned/deleted. The new moderation profile changes make up for it.
  - There is now a moderation profile link for moderators when user is banned/deleted
- added endpoints to view user profiles, view moderation profiles and update moderation profiles (for admin pages), also implemented into frontend
  - New feature: Moderators can now view account status, which includes network bans.
  - New feature: Moderators can view posts on the moderation profile
  - New feature: Moderation profile looks slightly different, so its easier to distinguish from normal profile.
  - New feature: Moderation profile now contain audotmods
  - New feature: User related reports can now show "resolved through similar report" as a resolution reason
  - Moderation profile is now split up into tabs
- Added endpoints for notifications, also implemented into frontend
- Deleted unused database methods
- Reworked how errors work in internal API, it now has error codes that can be handled on the frontend side
- Restyle deleted posts for web
- Reworked hamburger menu
  - improved styling with more reactive UX
  - Removed items that are irrelevant for the post
  - Fixed hamburger not visible when theres only one post on the page

<details>
  <summary>Pics of new moderation changes</summary>
image

---

image

---

image

---

image
</details>

Resolves a part of #112

PR #499, opened 2026-05-02
Improve moderator ban saving form

### LINKED ISSUE #260: [Bug]: "Null" Ban Expiry still acts as an infinite Ban Time
### Checked Existing

### What happened?

I am testing with an alt account what happens if a ban date is never put in for a limit from posting/temporary ban.

It acts like an infinite restriction. The notification for limit form posting says "invalid date" but my alt account is unable to post, so effectively permanently banned unless submits an appeal because of a moderator mistake (and a system bug).

In this case, I expected it to error out either on the moderator's end (to not submit a restriction) or to send a notification to the user while not actually restricting them as there is no actual date).

### Steps to reproduce?

1. On the Mod panel for a user in Juxt, select a Limit from posting or temp ban
2. do not input a date
3. enter

_No response_

mrjvs:
Changes:
- Improve moderator ban saving form
  - Rewrote all JS related to the form
  - Added a bunch of CSS for the form
  - Made it it's own component
  - Every account status now has it's own form, removed banLiftDate for normal status and permanent ban status
  - Made banLiftDate required for limitedFromPosting and temporaryBan

<details>
  <summary>Screenshots</summary>
    
image

---

image

</details>

Resolves #260 (Still allowed on backend, but I think thats fine since it's not a illegal data state)

Will conflict with #482 but easy to resolve as it's now it's own component


ashquarky (approved):
Looks good

PR #507, opened 2026-05-05
Implement config for ratelimiting

mrjvs:
Changes:
- implement config option for toggling ratelimits on juxtaposition-ui

With this, we'll be able to disable ratelimits in prod for a while to diagnose any issues. And then turn it back on when we're done.


ashquarky (approved):
Is that really the only endpoints we have ratelimits on? wowee

PR #512, opened 2026-05-07
Fixed community list missing communities

### LINKED ISSUE #511: [Bug]: "All communities" list only shows a maximum of 90
### Checked Existing

### What happened?

"All communities" list only shows a maximum of 90

"All communities" should show all communities

### Steps to reproduce?

1. Go to all communities on prod
2. Not all communities are there

This is likely introduced in the datasource refactors (#479 maybe)

mrjvs:
Changes:
- Added util to get all data from a paginated endpoint
- Used new util to get all communities on the "all communities" page

Resolves #511

### DIFF
--- a/apps/juxtaposition-ui/src/api/helpers.ts
+++ b/apps/juxtaposition-ui/src/api/helpers.ts
@@ -1,0 +1,31 @@
+export type AllFromListOptions = {
+	limit: number;
+};
+
+export type PageControls = {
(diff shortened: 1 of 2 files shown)


ashquarky (approved):
Kinda weird to do this instead of upping the limit or paginating the actual frontend, but I guess the implementation is fine even if I think the design is strange

Issue #334, opened 2026-02-08
[Bug]: CTR doesn't keep nagivation history well


ashquarky:
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What happened?

In CTR's "pjax lite" the URL history is just a list that's kept in a variable. Therefore, if we ever do a full page load, the history is lost and the toolbar "back" button behaves erratically.
There's a few ways to do page loads, but the most notable one I can think of is submitting a form (like making a post). Sometimes the console will also request a reload. Deleting a post also changes the href.
Probably this API can be moved to `lls` to get some persistence across page loads.

### What did you expect to happen?

pjax works good

PR #513, opened 2026-05-08
fix(ctr): use PJAX for form submissions

ashquarky:
Part of #334 

### Changes:

- Instead of submitting HTML forms normally, have them target an iframe and let PJAX do the page load. This keeps our client variables (critically, the nav history) intact.
- Since CTR gets very crashy if you load and then remove an iframe, put the iframe up in <body> where it stays permanently. It gets re-used between calls.
- Rework various endpoints that take form submissions to optionally return with "pjax api" - a HTML script tag that postMessages a JSON blob immediately. This is so that aforementioned iframe can communicate a tidy API response up to our bundle. Bundle includes status code, optional message to show the user, and href to load next. Message is alert()ed for now but could integrate with the new error display later.
- Since this touched all the error handling paths on the form submission endpoints, unify and tidy those up with an accept/reject function to ensure they behave consistently. This is a UX change and it sometimes shows full error pages in places it previously wouldn't, but I think being more honest is less confusing to the user.
- For posting, errors are shown through the new error display, from automod.
- For reporting, errors are shown as alert()s.
-  Improvements to PJAX Lite core. Attempts to actually handle XHR errors instead of locking users in a CTR transition, and less buggy history state logic.

- [x] I have tested all of my changes


ashquarky:
Mostly interested in whether the pjax_api idea is a decent one or whether these would be better as e.g. different URL endpoints

mrjvs (commented):
I think I want to explore a different direction first before deciding to go down the iframe route.
Especially with the iframe's side-effect of wanting to crash on unload, I feel like it's too fragile and too unconventional of a method of communication. I'd rather stick with more standard approaches to web-apps if possible.

I'd like to explore a method like this:
Client side:
1. Every `<CtxPjaxForm>` automatically adds a `<input type="hidden" name="__pageState" data-page-state />`
2. A JS script will populate every `*[data-page-state]` element when page state changes (like on view changes)
3. The window object has a `window.hydrate(pageState)` method, it replaces the page state with current.

Server side:
1. A middleware will detect a body param called `__pageState` and store it in the response locals
2. The `<CtrRoot>` grabs page state from the request adds this:
```ts
<InlineScript src={`window.hydrated(${JSON.stringify(pageState)})
```

And for popups or smth, we can probably make a new component that shows an error on render, just with standard data attributes.

This method will allow us to not have to modify our handlers so we don't have to do something specific for CTR, we can just write the endpoints as we normally would.

It may feel a little janky, but it was a common technique back in the day, it came default with aspx for example: https://learn.microsoft.com/en-us/archive/msdn-magazine/2003/february/cutting-edge-the-asp-net-view-state

ashquarky:
> This method will allow us to not have to modify our handlers so we don't have to do something specific for CTR, we can just write the endpoints as we normally would.

If I ported all the frontends to use an API-style call instead of a page load, would that be an improvement? On newer browsers we can do it directly with XHR instead of needing the iframe hack, but the server-side part would be the same (just parse the JSON blob directly)

Then it's consistent across frontends, it's just not a page load anymore. It'd be nice to do this for portal too since that brings it closer to being able to use PJAX Lite instead of full PJAX (and therefore get consistency across the frontends for pjax)

ashquarky:
> `Every <CtxPjaxForm> automatically adds a <input type="hidden" name="__pageState" data-page-state />`

My main concern with this idea is that the user's history stack is included in the page state here, so we can easily imagine several hundred URLs in a list being part of it. Maybe that'd be fine? I dunno how big the blob would be and especially how much parsing overhead that'd add onto the new page load (3DS CPU very sad)

mrjvs:
> My main concern with this idea is that the user's history stack is included in the page state here, so we can easily imagine several hundred URLs in a list being part of it. Maybe that'd be fine? I dunno how big the blob would be and especially how much parsing overhead that'd add onto the new page load (3DS CPU very sad)

Both chrome and firefox have a 50 item limit to the back button. So we can quite easily cap it at 50 or lower imo.
Plus we can safe space by just saving paths and not full urls.

ashquarky:
I guess that means we should revive frontend modules so we can have the state blob get accessed in a structured way

ashquarky:
I was looking at the applet-types and I think it'd be possible to polyfill `sessionStorage` across both platforms. Doing that would obviously be ideal since then we don't have to roundtrip to the server and hydration would be a lot cleaner. Will look into it

Issue #147, opened 2025-07-21
[Enhancement]: Improve infinite loading on Wii U/web and replace it with pagination on 3DS


superfever3000:
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What enhancement would you like to see?

On the Wii U and web browser versions of Juxtaposition, the infinite loading does not unload or do anything to unseen (not visible from browser view) posts, and since there's a lot of event handling, actions like giving posts a _Yeah!_ would either not work or be very slow until you refresh or go to the post itself.
For the 3DS version, I'd suggest doing away with it entirely and switching to traditional pagination like the original Miiverse. The 3DS is very underpowered so infinite loading will not work for it in the long run.

PR #519, opened 2026-05-13
feat(ctr): Pagination for post list views

ashquarky:
Part of #147 

### Changes:

- Restructures post feed lists. Instead of having a seperate /more endpoint, add an offset parameter to the main one. Factor out URL generation logic so it's consistent. Re-use post list view props across the board.
- Since the main endpoint now accepts offset and also knows how to render full pages (alongside partials), we get pagination for free! Add next and prev page links for the frontends to use.
- Use the next page link on CTR instead of infinite-scrolling. Replaces the .load-more button with one that's styled after the page buttons we recently added. This should dramatically improve RAM use and allow users to scroll feeds as much as they want without crashing the page.
- Fix topics feed including removed posts and replies, oops!
- Add indexes for various common feed queries

### DIFF
--- a/apps/juxtaposition-ui/src/models/post.js
+++ b/apps/juxtaposition-ui/src/models/post.js
@@ -112,6 +112,14 @@ PostSchema.index({
 	parent: 1
 });
 
+// Topic tag lookups
+PostSchema.index({
+	topic_tag: 1,
+	parent: 1,
+	removed: 1,
+	created_at: -1
+});
+
 PostSchema.methods.upReply = async function () {
 	const replyCount = this.get('reply_count');
 	if (replyCount + 1 < 0) {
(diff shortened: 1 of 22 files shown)


mrjvs (approved):
I kind of wonder if we want to have a previous page button for ctr. Since it now replaces the whole page list, users now have no way of going back up.

Looks good though!

ashquarky:
> I kind of wonder if we want to have a previous page button for ctr. Since it now replaces the whole page list, users now have no way of going back up.
> 
> Looks good though!

The idea was they could still use the browser chrome (other PRs fix up PJAX to make it reliable enough for this). But yeah, I'd be down to add a proper forward/back swticher, maybe something sticky, just need to work out a design.

suprstarrd:
Cheered so loud at this title

ashquarky:
> Cheered so loud at this title

It's in dev if you wanna play ^^

imora7024:
Personally from testing it. a bit... Im really satisfied with it. 
congrats on the merge! 
It works almost identically (if not the same) as it was back on both Rverse and OG Miiverse!

PR #520, opened 2026-05-14
hotfix(web): Make guest check reflect hasAuth() method

ashquarky:
Resolves #XXX

**Hotfix:** This PR is against master, not dev!

### Changes:

We are receiving reports that users are seeing 404 screens and other issues when their tokens expire. The UI has a system where it extends token lifetimes by caching the user data blob in a cookie for read access, but the new API backend doesn't do this. This leads to weird behaviour when the UI thinks a user is logged in but the backend doesn't.
Fix this by adjusting the UI's guest access check to align with what the hasAuth() helper actually does. This will make users sign in again when the backend rejects their token.

This will mean that users don't get the benefit of the caching system and will have to log in hourly - this whole system desperately needs a refactor and refresh token auth, but I'm keeping the change minimal so we can hotfix it straight into prod and do a more proper tidy up later.

Same commit is available against dev in the `patch/guest-access` branch if that is preferable.

### DIFF
--- a/apps/juxtaposition-ui/src/middleware/webAuth.ts
+++ b/apps/juxtaposition-ui/src/middleware/webAuth.ts
@@ -45,7 +45,7 @@ export const webAuth: RequestHandler = async (request, response, next) => {
 	}
 
 	// Handle guest access pages
-	if (!request.pid) {
+	if (!request.pid || !request.user || !request.self) {
 		if (!requestOkForGuest(request)) {
 			return loginWall(request, response);
 		}


ashquarky:
A sample endpoint with bad behaviour is `/feed`. If you log in, edit your token in DevTools to be garbage, then look at /feed, you'll get a 404 page, and then going to the homepage takes you to firstrun instead of asking you to log in (!).

mrjvs (approved):

PR #523, opened 2026-05-17
fix(web): Use WebCommunityIcon everywhere, improve community grid

ashquarky:
Resolves #XXX

### Changes:

After #410 landed, the API has been very kindly finding a valid icon for communities, even ones that don't have native ones (-> some subcommunities). This change ports all community icons in web to use the WebCommunityIcon component, which automatically pulls that data. This means those communities have their correct icons displayed. This is particularly noticeable in the new Related Communities view.

As usual in Juxt Web, touching something made the layout explode, so I also redid the elements in the community grid to use flexbox, and various list views to use one big anchor tag rather than lots of small ones. I had to keep the div structure the same in the latter though to avoid breaking the CSS - the whole list component needs to be rebuilt, but the goal of this PR was just to fix up the Related Communities view.

CommunityViewData had to go due to not including the icon fields, though since the caller was fetching the full community object anyway it's no great loss. 

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -23,4 +23,7 @@
     "[typescriptreact]": {
         "editor.defaultFormatter": "dbaeumer.vscode-eslint"
     },
+    "[scss]": {
+        "editor.defaultFormatter": "vscode.css-language-features"
+    },
 }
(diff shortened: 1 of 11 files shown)


mrjvs (approved):
Gave it a quick test and it looks to be styled the same. lgtm!

Issue #502, opened 2026-05-03
[Enhancement]: Enhancements idea for Juxtaposition off-device


nebnebgamez:
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What enhancement would you like to see?

Hello team, this is my first issue so if anything is messed up, I hope you guys don't mind.

The reason I'm starting this is because I have some ideas to improve the design of Juxt offdevice. Some of these include
1. Not having the all communites button on /titles touch the button of the webpage - This isn't great when trying to maintain a clean and modern design, and fixing it within the future would be nice
2. Unifying the titles for each page - I noticed that on only some pages, it uses a title like "Juxt - nebneb" while other pages just use the URL, having all pages unified to have a title would also be nice
3. Timestamp under the notification description (or whatever you call it, the stuff that says that someone has followed you and stuff) - This is more of a personal idea but it'd honestly be a clean design choice
4. Fix name/timestamp + post body alignment - I noticed that post-meta-wrapper uses a margin-left of .5em which is making it not align with post body, I don't know if this is intentional but it's not really a clean design choice. I was able to fix it in inspect element by changing it to 1.1em

nebnebgamez:
Number 5 can be disregarded as it may not fit the category that this issue is in, but the others would be nice for improving design of Juxt offdevice

nebnebgamez:
7 and 6 could also as they're not 100% needed or anything and it's just ideas

ashquarky:
These are all reasonable design refinements - @mrjvs do you think a large-scale overhaul is coming quickly or should we open this up for contributons in the meantime

mrjvs:
A design-overhaul is not gonna be there for a while, so these types of contributions should be fine.

The current list of items is fine. but the list was edited after creation. Before the edit it had stuff that wasn't design-related and included items that would need more discussion.

ashquarky:
Approving list as it appears today, welcome for external contributors

PR #524, opened 2026-05-17
improve responsiveness by removing janky and hardcoded CSS

mrjvs:
Changes:
- improve responsiveness by removing janky and hardcoded CSS
- Renamed `#wrapper` selector, since it wasn't a unique element.

Mostly just removed all the width hardcodings, it now relies on the wrapper element for which size to take.

I removed automatic align-items center from all web wrappers, since it caused *many* issues with responsiveness. Because of that I was able to remove max-width basically everywhere.

While I tried to test everything myself, it will likely need some extra clicking around by someone else.

Resolves a part of #502 (nothing touches bottom of pages now)


apps/juxtaposition-ui/webfiles/web/css/web.scss
```diff
@@ -190,17 +190,40 @@ header > .selected svg {
...
+  // Page padding
+  &.page {
+    margin-top: 0rem;
+    margin-bottom: 4rem;
+  }
+  
+  // Only have bottom padding
+  &.bottom {
+    margin-bottom: 4rem;
+    margin-top: 0;
```
ashquarky, line 211: I'm not actually sure this is needed due to margin collapsing (top/bottom margins overlap by default) but it's understandable to not rely on that

ashquarky (commented):
The "default" page is 600px flat instead of 600px+2em, which is juuust too thin for 5 buttons in English now

Image

"Accounts" and "Communities" tabs in mod panel broke. Note search bar also moved, and on Communities tab, the Create Community tab got very wide.

Image

Otherwise really good

apps/juxtaposition-ui/webfiles/web/css/web.scss
mrjvs: I tend to find margin collapse to not be great to rely on, it crosses component boundaries and makes it difficult to understand what's going on. (plus it can randomly disappear with seemingly random changes like using flexbox)

mrjvs:
Ill fix the bugs you found tomorrow 👍

mrjvs:
Fixed all the bugs I think, ready for another review

ashquarky (dismissed):
LGTM other than merge conflicts (sorry)

ashquarky (approved):

Issue #277, opened 2026-01-02
[Enhancement]: Better community discovery


superfever3000:
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What enhancement would you like to see?

Currently, Juxtaposition has over 60 communities that are visible in the "All Communities" page, the current way of community discovery is not viable for a large amount of communities.
Right now, the way to discover communities is by clicking/tapping on "All Communities" and scrolling through a long list with no organization or distinction.
When Juxtaposition was released to the public back in December 2023, it had a small number of communities, and many of those were Wii U games, which justified the original community organization.
As Juxtaposition grew, so did the number of communities, including more 3DS communities than ever before, making it harder to discover a particular community, especially for the user's respective console. This is why I created this issue because if there isn't an update to discovery, than the problem will only grow stronger.

PR #525, opened 2026-05-19
feat(ui): Add client-side searchbar to the all communities page

ashquarky:
Part of #277 

### Changes:

Adds a search bar to the top of the All Communities page. For now, it just shows and hides the elements on the client side. If we ever add pagination to this then the search feature has to go server-side; but this was quick to implement and in high demand by the userbase, especially as we've been adding lots of communities lately.

image
image

The search algorithm used can deal with colons, capital letters, spacing differences and special characters:

image

There's a special case for Pokémon:

image

On 3DS and WiiU, we don't get change, input or beforeinput events on text fields, so I had to fall back to a setInterval based approach. It cancels the interval during a PJAX load so the memory and CPU usage doesn't stack up.


ashquarky:
Inspired by a Juxt post I can't find now in which a user begged us to please add a Mario Kart 7 community

apps/juxtaposition-ui/webfiles/web/js/components/ui/WebSearchForm.ts
```diff
@@ -0,0 +1,26 @@
+function prepSearchTerm(term: string): string {
+	const removeChars = /\W/g;
+	// é for Pokémon
+	return term.toLowerCase().replace('é', 'e').replace(removeChars, '');
+}
```
jonbarrow, line 7: This isn't super helpful for the console clients, but on the browser at least we can do actual normalization. For example, for full-string comparisons [`Intl.Collator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator) is probably the best:

```js
const collator = new Intl.Collator('en', {
    sensitivity: 'base'
});

console.log(collator.compare('pokemon', 'Pokémon') === 0);
```

But if there's a need to match substrings, we can use the [`normalize`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) API:

```js
// * Decomposes the characters and removes combining diacritical marks
// * https://en.wikipedia.org/wiki/Combining_Diacritical_Marks
'pokémon'.normalize('NFD').replace(/[\u0300-\u036f]/g, ''); // * Turns into "pokemon"
```

These are somewhat modern features though. Intl.Collator didn't land in Chrome until version 24 and the normalize API didn't land in Chrome until version 34. So these may not be in the console clients, but it may still be worth checking

Alternatively, the search could be moved to the server like you already mentioned where we know we have access to these APIs. I doubt Pokemon is going to be the only time this issue crops up, so having a more general normalization step will likely help in the long run

mrjvs (dismissed):
Maybe im misjudging the problem, but it looks like it will be easier to do on the (juxt-ui) server side:
- You have access to fancy APIs and libs (like Jon's suggestion)
- You have all the data and can easily just .filter() the community list
- with input.onchange you can just load in the html and replace the body, with a small debounce.

I think both approaches are fine, although we do want it on the api end in the long term.

Since im fine with both approaches, I'm gonna give this an approval. Feel free to merge if you prefer the client-side approach

ashquarky:
I'm wanting to go with client-side mostly since adding all the endpoints and database queries is something I don't feel like doing right now, though I acknowledge it'll need to happen eventually for pagination. ^^

As to the Intl. API point, I hope this is the best of both worlds!

mrjvs (approved):

Issue #15, opened 2025-02-25
[Enhancement]: Switch to Postgres


mrjvs:
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What enhancement would you like to see?

Mongodb has it's issues and it's starting to show on Juxt.

It's time to switch to switch to PostgreSQL.

We will be using the Prisma ORM, used only on the new `Juxt-API` project.

jonbarrow:
> Mongodb has it's issues and it's starting to show on Juxt.

Can you provide any examples of this which aren't user error? Mongo has been working just fine in arguably higher traffic servers like the account server, and nearly every issue we've had with Mongo has been user error or some other issue unrelated to the database itself

jonbarrow:
Details were added since my last comment. That sounds reasonable to me

ashquarky:
Since this will involve schema design and changes, we should consider our design to allow extra moderation logging and automation later on - see #102

PR #557, opened 2026-07-02
Rework notifications

### LINKED ISSUE #401: [Enhancement]: Handle notification content as an enum
### Checked Existing

Notifications currently store a couple things in the db:
- type
- text
- users
- image
- link

When type is `notice`. The text is the DB is displayed with the image and a link.
This is very inflexible and has caused trouble in prs like #391. Localization is also impossible for the notification text because of this system.

What it should be:
- Type `notice` should have subtypes for the different kinds of notices
- The subtype will determine the text, link and image. And is done on render, not on notification save
- There should be one subtype: `custom` that will allow dynamic text to be displayed, this is for edgecases like sending specific messages to users.

_No response_

mrjvs:
Changes:
- Install prisma ORM with postgres
- Update development setup to support postgres
- Rewrote local development docs
- Remove mongoose model for notifications
- Create postgres models for notifications
  - NotificationRecipient: user who receives the notification
  - Notification: content of a notification
- Port all notification code to prisma
- Added two more notification types: PostDeleted and LimitedFromPosting
  - We want to every notification to be a separate type, so they are translatable and can support custom UI
- Added translations for the new notification types
- Removed Github Actions for bundle analysis, it's broken
- Add data migration to port the notifications from mongodb to postgres and detect the types properly

Notes:
- Infra will need to be updated to support postgres for juxt
- Infra also needs an init container to run migrations
- I've manually checked a lot of the migrated notifications, they all look correct (Even the ones with wacky data, the wacky data was in the DB like that)

Resolves #401 
Part of #15


mrjvs:
@Joshua8600 if you want to help out. here is a fun challenge:

In `/migrations/8-notification-to-pg/script.ts` there is a function called `parseDateString(str)` and it's currently unimplemented.

The input of the method will be any of these 4 formats (all of them strings):
```md
# Format 1
Thu Jan 01 1970 00:00:00 GMT+0000
Sat Oct 18 2025 08:40:00 GMT+0000
Wed Dec 31 2025 16:31:00 GMT+0000

# Format 2
# In between 0:00 and PM/AM there is a U+00a0 character
Jan 4, 2026, 6:40 PM UTC
Dec 31, 1, 11:49 PM UTC
Jan 1, 10000, 12:00 AM UTC
May 21, 2026, 12:30 PM UTC
Jan 1, 2101, 5:59 AM UTC
Jun 15, 2026, 5:00 PM UTC

# Format 3
null

# Format 4
Invalid date
```
Format 1 and 2 need to parsed into a JS Date object.
For format 3 and 4, you can just fall back to `new Date()`

For testing, I recommend making a new file (`test.js`), placing your JS code in there and then run `node test.js` in your terminal to run it.

You can base the PR off of `feat/notifications-rework` and merge it in there as well.

Joshua8600:
PR made: #561

apps/juxtaposition-ui/src/services/juxt-web/views/ctr/notificationListView.tsx
```diff
@@ -10,60 +10,150 @@ import type { NotificationItemProps, NotificationListViewProps, NotificationWrap
...
+							</span>
+							<span className="timestamp">
+								{' '}
+								{humanFromNow(data.updatedAt)}
+							</span>
+						</p>
+					</a>
+				</div>
+			</>
+		);
```
binaryoverload: Kinda feels like these cases should be split into components so you can make the top level logic concise - e.g.

```js
switch (type) {
  "limitedFromPosting":
    return <LimitedFromPosting data={data} />
etc
```

apps/juxtaposition-ui/src/services/juxt-web/views/web/notificationListView.tsx
binaryoverload, line 1: Is it possible for the views to share components?

 I guess we won't be adding many types (I expect) so it's unlikely to become a problem but it would neaten things up

binaryoverload (changes requested):

apps/miiverse-api/eslint.config.mjs
```diff
@@ -8,5 +8,8 @@ export default defineConfig([
 			// Zod has too complex types to make a return type
 			'@typescript-eslint/explicit-function-return-type': 'off'
 		}
+	},
+	{
+		ignores: ['src/prisma']
```
binaryoverload, line 13: What is this ignoring? There is no src/prisma

binaryoverload (changes requested):

apps/miiverse-api/eslint.config.mjs
mrjvs: It's generated code by prisma. It doesn't exist in commit history but it will exist on the devs machine.

apps/juxtaposition-ui/src/services/juxt-web/views/web/notificationListView.tsx
mrjvs: Quarky and I decided early on that we don't share components unless it's completely abstracted away from markup (like `<T />`). There's too many edgecases between platforms and we don't want to build into components.

apps/juxtaposition-ui/src/services/juxt-web/views/ctr/notificationListView.tsx
mrjvs: Took a bit, but should be good now

binaryoverload (approved):

binaryoverload (changes requested):
Not to merge until dev environment is setup for postgres

binaryoverload (approved):

PR #567, opened 2026-07-07
Added Succesful Report Toast

### LINKED ISSUE #454: [Enhancement]: Add "Report Sent" pop-up
### Checked Existing

There is a pop-up notification when a link from a post is copied, but nothing is displayed when you report someone.

It would be good to have this so that you know whether the report was actually sent or not.

_No response_

DevDrew64:
Resolves #454 

### Changes:

**Summary:** Adds a toast notifcation confirming successful post reports, replacing the full-page redirect on report submission.

**Problem:** Previously, submitting a report via the report modal triggered a native HTML form POST which caused the entire page to reload, giving no feedback to the user that their report was successful.

**Modifcations:**
1. New file: apps/juxtaposition-ui/webfiles/web/js/toast.ts
Extracted the Toast, initToast, and startHideToast functions out of web.js into their own module so they can be imported and reused by other client-side modules (previously they were bare, non-exported function declarations only usable within web.js itself).

2. Modified web.js to remove the local Toast and initToast function declarations as I moved them to toast.ts.

3. Modified reports.js to invlude the moved toast logic and trigger it when a report is made.

4. Modified apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.tsx to handle reports without triggering a refresh of the page.

- [ ✅ ] I have read and agreed to the [Code of Conduct](https://github.com/PretendoNetwork/Pretendo/blob/master/.github/CODE_OF_CONDUCT.md).
- [ ✅ ] I have read and complied with the [contributing guidelines](https://github.com/PretendoNetwork/Pretendo/blob/master/.github/CONTRIBUTING.md).
- [✅  ] What I'm implementing was an [approved issue](../issues?q=is%3Aopen+is%3Aissue+label%3Aapproved).
- [ ✅ ] I have tested all of my changes.


apps/juxtaposition-ui/webfiles/web/js/reports.js
mrjvs, line 51: A side-effect of not reloading the page is that the value of the "extra notes" section of the modal persists across multiple reports.

Let's clear the input field in `reportPost()` to resolve that.

apps/juxtaposition-ui/src/services/juxt-web/routes/console/posts.tsx
```diff
@@ -296,7 +296,7 @@ postsRouter.post('/:post_id/report', upload.none(), async function (req, res) {
 		reasonId: body.reason
 	});
 
-	return res.redirect(`/posts/${post.id}`);
+	res.sendStatus(200);
```
mrjvs: 3DS and WiiU still rely on the redirect to still work.

We don't really have a standardized way to deal with that (yet), so let's just add a new query parameter: `?api=true`. If it's true then status codes can be returned rather than redirects.

apps/juxtaposition-ui/webfiles/web/js/toast.ts
mrjvs, line 1: Good idea to move it to it's own file :+1:

Could also update `login.js` to reference this file? It had the Toast logic copied over.

mrjvs (changes requested):
Thanks for the PR!

Found a couple bugs in testing, but nothing that’s hard to fix :D

mrjvs (changes requested):
Looks good, just linting left 👍 

(tip: You can run auto-fix linting by running `npm run lint:fix` in the project folder)

mrjvs (approved):

mrjvs:
Thank you for your contribution! We're doing a release tomorrow, so your changes are gonna be in prod nice and quick.

PR #569, opened 2026-07-08
Rework user table

### LINKED ISSUE #73: [Bug]: Banned followers still show up in followers list
### Checked Existing

- [X] I have checked the repository for duplicate issues.

### What happened?

Apparently, if someone follows someone else, and the one who follows gets banned, the followed user can still see the banned user in their followers
Dunno if this is intentional

Either the name to be hidden or the follow to be removed/hidden entirely on ban

### Steps to reproduce?

1. Have an user follow you
2. Have that user get banned 
3. Still being able to see the user in your follower list

_No response_

mrjvs:
Changes:
- Replace `Settings` entity with two tables: `users` and `user_settings`. Removed `Settings`
- Move the follower system from `Content` entity to a dedicated `user_follows` table. Removed `Content`
- Update all miiverse-api follower code to use the new table
- Update all miiverse-api code to use the new user tables
- Update metrics to use the new tables instead of the old ones

Notes:
- Olive API has had some slight changes:
  - Having user content is no longer a requirement, only requires a User table where it checks for it
  - Follower logic has changed, it seems it was bugged but it needs some verifying
- There are a couple of broken followers, the data migration simply skips them. Otherwise you get foreign key constraint errors.
- For next PR: Update ban system so it's a dedicated table with history and not one field on the User

Testing:
- Test miiverse functionality

Part of #15 
Fixes #73


mrjvs:
After review, don't merge yet. A release is queued.

mrjvs:
This is ready for review.
The only thing that's been untested is the miiverse functionality. I've had to change the logic a little bit, I also think it was just plain wrong the way it was implemented originally.

apps/miiverse-api/src/models/migrations/20260708143652_users/migration.sql
```diff
@@ -0,0 +1,63 @@
+-- CreateEnum
+CREATE TYPE "ProfilePrivacyType" AS ENUM ('Public', 'UsersOnly');
```
ashquarky, line 2: Haven't used a SQL enum before - We can add extra options later without stuffing it yeah?

apps/miiverse-api/src/models/migrations/20260708143652_users/migration.sql
```diff
@@ -0,0 +1,63 @@
...
+
+-- CreateTable
+CREATE TABLE "users" (
+    "pid" INTEGER NOT NULL,
```
ashquarky, line 6: Worth noting pid is a uint32 in the API, so we could possibly get out-of-range values here.
https://github.com/PretendoNetwork/grpc/blob/465a477acea025846e04ff35e84697b124a82d98/protobufs/account/v2/exchange_token_for_user_data_rpc.proto#L14

Today's account server is hard-coded to never go anywhere near this though, so it's probably fine :tm: 
https://github.com/PretendoNetwork/account/blob/edb4b02e9fe99d6dd22cfdd9e86b43e71d7c57cd/src/models/pnid.ts#L152

apps/miiverse-api/src/services/api/routes/posts.ts
```diff
@@ -373,7 +372,7 @@ async function newPost(request: express.Request, response: express.Response): Pr
 		id: '', // * This gets changed when saving the document for the first time
 		title_id: request.paramPack.title_id,
 		community_id: community.olive_community_id,
-		screen_name: userSettings.screen_name,
+		screen_name: dbUser.displayName,
```
ashquarky, line 375: I believe we have `request.user` here fresh from Accounts with the same data, it may be worth considering removing the displayName column entirely?

ashquarky: Doing that would mean we spam Accounts with user data requests tho

ashquarky (approved):
Looks clean

apps/miiverse-api/src/models/migrations/20260708143652_users/migration.sql
mrjvs: I haven't used them much either, but according to docs adding enum values is supported from postgres 9 onwards.

The migrations for it will be generated by prisma and prisma supports doing it. So I trust that itll be fine.
We likely will have to write some custom SQL migrations if we want to remove an option. but we'd need to do that either way, doesnt matter much if its SQL or JS scripts.

apps/miiverse-api/src/models/migrations/20260708143652_users/migration.sql
mrjvs: Unfortunately postgres doesnt support unsigned ints. And if we go into the int64 numbers then Javascript will have to start using bigints. And those are a very big pain in the ass cuz they aren't numbers, that means toString() and fromString() everywhere.

I think we can just trust the account server to not give invalid pids.

apps/miiverse-api/src/services/api/routes/posts.ts
mrjvs: I actually think keeping a cached variant is a good thing.

Especially if you have to fetch a list of 50 posts. I'm fine with a join to the user table, but making 50 requests to account for just a display name is a bit extreme.

If we keep the cached display name on the user and sync the cache every request the user makes, we no longer have the problem of posts having old display names (join the user table instead of on the post itself). and it doesn't spam account server more if you need to request large amounts of posts.

mrjvs: The only possibility for a desync is when the user updates their mii, but never visits juxt again.
Which I think is an acceptable downside to the system.

ashquarky:
LGTM then :+1:

PR #572, opened 2026-07-18
GRPC token update

### LINKED ISSUE #111: [Enhancement]: Use account GRPC for parsing tokens
### Checked Existing

Currently, juxtaposition-ui is decrypting the auth tokens to get the PID. Then uses that to fetch user data.

This should be changed to use the account server GRPC methods to get user data in one go. No decryption necessary.

This should probably be done after/while moving datasource to miiverse-api. Otherwise effort is wasted.

mrjvs:
Changes:
- Update as much as possible to use V2 GRPC methods, with a couple exceptions:
  - `getPNID()` on v2 account is not implemented. So it's using the deprecated `getUserData()`
  - Friends is hard to selfhost locally because of the new encryption keys. So it's gonna stay on v1 so local testing is possible.
- Replaced decrypting service tokens with the new `exchangeIndependentServiceTokenForUserData()`
- Removed unused packages related to decrypting tokens
- Removed unused types related to decrypting tokens

Notes:
- Web has been tested. So that's all the GRPC updates
- Service tokens have not been tested, I don't have a console set up. Hoping to get some staff to test it on dev.

Resolves currently broken production issues due to updated token format.

resolves #111

### DIFF
--- a/apps/miiverse-api/src/util.ts
+++ b/apps/miiverse-api/src/util.ts
@@ -3,12 +3,12 @@ import { DeleteObjectsCommand, PutObjectCommand } from '@aws-sdk/client-s3';
 import { SystemType } from '@pretendonetwork/grpc/account/v2/token_info';
 import { config } from '@/config';
 import { logger } from '@/logger';
-import { grpcAccount, grpcApi, grpcFriends } from '@/grpc';
+import { grpcAccount, grpcApi, oldGrpcFriends } from '@/grpc';
 import { getS3 } from '@/s3';
 import { AutomodLog } from '@/models/automodLog';
 import type { IncomingHttpHeaders } from 'node:http';
 import type { ObjectCannedACL } from '@aws-sdk/client-s3';
-import type { FriendRequest } from '@pretendonetwork/grpc/friends/v2/friend_request';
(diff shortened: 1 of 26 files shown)


ashquarky (dismissed):
LGTM

ashquarky:
Resolves #111

apps/miiverse-api/src/util.ts
```diff
@@ -60,72 +57,42 @@ export function decodeParamPack(paramPack: string): ParamPack {
 	};
 }
 
-export function getPIDFromServiceToken(token: string): number | null {
+export async function getUserDataFromServiceToken(token: string): Promise<GetPNIDResponse | null> {
 	try {
-		const decryptedToken = decryptToken(Buffer.from(token, 'base64'));
+		const userData = await grpcAccount.client().exchangeIndependentServiceTokenForUserData({
+			token
```
ashquarky, line 63: Missing API key here too

ashquarky (changes requested):

apps/miiverse-api/src/util.ts
mrjvs: Miiverse-api has auto-added headers for grpc clients. See /apps/miiverse-api/src/grpc.ts

ashquarky (approved):

PR #580, opened 2026-08-12
feat(ctr): List view fixes and community platform bubbles

ashquarky:
Part of #458 

### Changes:

Rebuilt the venerable `.list-content-with-icon-column` class, taking advantage of webkit box (proto-flexbox) and tidying up the alignment.

The borders and margins are now consistently aligned on lists, along with accurate text centering. The last item in the list also no longer has a border.

The selection box perfectly wraps the content instead of including parts of the margins and borders.

Community listings now include coloured dots to indicate the community's target platform.

Still to do:
- Work out wtf is going on with type="icon" icons - candidate for refactoring out entirely

- [x] I have read and agreed to the [Code of Conduct](https://github.com/PretendoNetwork/Pretendo/blob/master/.github/CODE_OF_CONDUCT.md).
- [x] I have read and complied with the [contributing guidelines](https://github.com/PretendoNetwork/Pretendo/blob/master/.github/CONTRIBUTING.md).
- [x] What I'm implementing was an [approved issue](../issues?q=is%3Aopen+is%3Aissue+label%3Aapproved).
- [x] I have tested all of my changes.

PR #585, opened 2026-08-19
feat: Add new notification types (yeahs, replies) and user settings to enable/disable them

ashquarky:
Resolves #XXX

### Changes:

Adds new notification types (in addition to the existing Follow notifications) - Reply and Empathy (Yeah!). Empathy notifications are grouped like Follows are, while Reply notifications are 1:1. 

All notifications also have user settings to turn delivery on or off to prevent users getting overwhelmed with notifications.

Depends on #580 


apps/miiverse-api/src/services/internal/utils/notifications.ts
```diff
@@ -112,21 +145,89 @@ export async function createNewFollowNotification(db: PrismaClient, ops: FollowN
...
+			pid: targetPid,
+			notification: {
+				type: 'Reply',
+				updatedAt: {
+					gte: new Date(now.getTime() - weekInMs) // Get 7 days worth of notifications
+				},
+				content: {
+					path: ['parent'],
+					equals: post.parent!
+				}
```
mrjvs, line 203: I prefer asserting (`if (!post.parent) throw ...`) instead of type casting. Type casting can lead to some funky behaviour if it's ever not what you expect it to be, in this case it can query the wrong data instead of just erroring out.

apps/miiverse-api/src/services/internal/utils/notifications.ts
```diff
@@ -43,57 +68,65 @@ export type LimitedFromPostingNotificationContent = {
...
 			notification: {
-				type: 'Follow',
+				type,
 				updatedAt: {
-					gte: new Date(now.getTime() - weekInMs) // Get 7 days worth of follow notifications
+					gte: new Date(now.getTime() - weekInMs) // Get 7 days worth of notifications
+				},
+				content: {
+					path: ['post'],
+					equals: empathyPost
```
mrjvs: Does this still work with followers? That type doesnt have a post.

apps/miiverse-api/src/services/internal/utils/notifications.ts
```diff
@@ -112,21 +145,89 @@ export async function createNewFollowNotification(db: PrismaClient, ops: FollowN
...
+			}
+		},
+		include: {
+			notification: true
+		}
+	});
+	const notifsContent = recentNotifs.map(v => v.notification.content as ReplyNotificationContent);
+	const hasNotifContentForUser = notifsContent.some(content => content.pid === post.pid);
+	if (hasNotifContentForUser) {
+		// Don't send any notification to prevent follow notif spam
```
mrjvs, line 213: correction: `[...] to prevent notif spam [..]`

mrjvs (changes requested):
Works nicely after a round of testing.

Just some nitpicks, and can then be merged

mrjvs (approved):

PR #595, opened 2026-08-31
Feat: gRPC API

CaramelKat:
Resolves #XXX

### Changes:

Marking this as a draft for now b/c I think that I may need to go and tweak the proto a bit after implementing everything, and I haven't done any testing yet aside from running the linter/compiling. 

This adds support for a number of grpc methods to pull data from Juxt, mirroring (most) of what is available over the xml console API. 

### DIFF
--- a/apps/miiverse-api/src/models/community.ts
+++ b/apps/miiverse-api/src/models/community.ts
@@ -206,7 +206,7 @@ CommunitySchema.method<HydratedCommunityDocument>('proto', async function proto(
 			96: this.icon_paths?.[96] ?? '',
 			128: this.icon_paths?.[128] ?? ''
 		},
-		pid: this.owner ?? 0, // TODO: This probably should have been made optional, oops
+		pid: this.owner ?? 0,
 		appData: this.app_data,
 		isUserCommunity: !!this.owner, // We assume a community is created by a user if it has an owner
 		createdAt: this.created_at.toISOString(),
--- a/apps/miiverse-api/src/models/community.ts
+++ b/apps/miiverse-api/src/models/community.ts
@@ -2,6 +2,8 @@ import crypto from 'node:crypto';
 import { Schema, model } from 'mongoose';
 import { MongoError } from 'mongodb';
 import { CommunityShotModes } from '@/types/mongoose/community';
+import { getUserAccountData } from '@/util';
+import type { CommunityData as ProtoCommunityData } from '@pretendonetwork/grpc/miiverse/v2/community';
 import type { ICommunity, ICommunityMethods, CommunityModel, ICommunityPermissions, HydratedCommunityDocument, ICommunityInput, IIconPaths } from '@/types/mongoose/community';
 import type { CommunityData } from '@/types/miiverse/community';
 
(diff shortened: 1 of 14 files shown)


apps/miiverse-api/src/models/community.ts
```diff
@@ -192,6 +194,42 @@ CommunitySchema.method<HydratedCommunityDocument>('json', function json(): Commu
 	};
 });
 
+CommunitySchema.method<HydratedCommunityDocument>('proto', async function proto(): Promise<ProtoCommunityData> {
```
jonbarrow, line 197: I would probably suggest either changing this `proto` name to something more descriptive or adding TSDoc comments, so it's more clear in places like intellisense what the method is used for

apps/miiverse-api/src/services/grpc/getCommunityPosts.ts
```diff
@@ -0,0 +1,103 @@
...
+		case PostType.POST_TYPE_TEXT:
+			query.body = { $nin: [null, ''] };
+			break;
+		case PostType.POST_TYPE_MEMO:
+			query.painting = { $nin: [null, ''] };
+			break;
+		case PostType.POST_TYPE_SCREENSHOT:
+			query.screenshot = { $nin: [null, ''] };
+			break;
+	}
```
jonbarrow, line 58: I may have missed this in the gRPC PR since this data comes from there it seems, but posts only have 2 types not 3. They can either be `text` which contains body text and no painting, or `memo` which contains a painting and no body text. Both types can have a screenshot though, that isn't a separate type

apps/miiverse-api/src/services/grpc/getUserPosts.ts
```diff
@@ -0,0 +1,64 @@
...
+		case PostType.POST_TYPE_TEXT:
+			query.body = { $nin: [null, ''] };
+			break;
+		case PostType.POST_TYPE_MEMO:
+			query.painting = { $nin: [null, ''] };
+			break;
+		case PostType.POST_TYPE_SCREENSHOT:
+			query.screenshot = { $nin: [null, ''] };
+			break;
+	}
```
jonbarrow, line 46: Same post type comment as above

jonbarrow (commented):
Lookin good chief! I'm sick right now so I just gave this a quick scan, only a couple non-major things/questions from me rn

apps/miiverse-api/src/services/grpc/getCommunityPosts.ts
CaramelKat: I know either can have screenshots, the intention was just to offer the ability to request them so if a future integration only wanted posts that have screenshots, they could do that. It felt weird to allow you to do that for text and paintings, but not screenshots imo

apps/miiverse-api/src/services/grpc/getPost.ts
```diff
@@ -0,0 +1,20 @@
+import { ServerError, Status } from 'nice-grpc';
+import { Post } from '@/models/post';
+import type { GetPostRequest, GetPostResponse } from '@pretendonetwork/grpc/miiverse/v2/get_post_rpc';
+
+export async function getPost(req: GetPostRequest): Promise<GetPostResponse> {
+	const post = await Post.findOne({
+		id: req.postId
+	});
```
mrjvs, line 9: Make sure to filter out deleted posts from here

mrjvs: Same thing with `getPostReplies`, you shouldn't be able to get replies from deleted posts

mrjvs (changes requested):

apps/miiverse-api/src/services/grpc/getPost.ts
CaramelKat: Both should be good now, thanks for catching that

mrjvs (approved):
Looks good.

If you still want to give it a couple tests. I think the missing step you were running into is running `npm run migration:deploy` before starting miiverse-api.

It's a relatively new step, so idk if it's been documented yet.

(Also sorry about the delay, I had this approval message in draft on accident, never submitted ;-;)

PR #604, opened 2026-09-13
feat(webfiles): Navigation fixes for CTR and Portal, new common library

### LINKED ISSUE #334: [Bug]: CTR doesn't keep nagivation history well

### LINKED ISSUE #475: [Bug]: Close button shown after a post is reported.
### Checked Existing

### What happened?

When reporting a user (today, 2026-04-07), after submitting the report the "Go Back" button changed to the general Close button. Despite this, pressing the B button (without touching 'Close') did allow me to go back, but then showed the Report page again before being able to back out into the post and finally the list of posts.

The button remains "Go Back", and optionally the history isn't kept on the report page and coming back to the post page, so pressing B just returns you to the previous page (typically, the community page)

### Steps to reproduce?

1. Launch Juxtaposition on Wii U.
2. Select a game community.
3. View a post to be reported.
4. Press report post, submit a report.
5. Observe Go Back button -> Close button.

_No response_

ashquarky:
Resolves #334
Resolves #475 

### Changes:

- Fixes import aliases and tsconfig for the frontend JS.
- Adds a "common" folder for the three frontend's JS and moves shared code (XHR, api) into it.
  - Code in these folders should use the lowest common denominator across the three platforms. Thanks to ctr's polyfills, this really isn't too bad.
  - Importing this requires a mildly cursed TS setup where there's four (!) overlapping tsconfigs, but it seems to work, so!
- Introduce a wrapper for localStorage/sessionStorage that works across frontends.
  - This uses a bundler define to hide platform-specific code, so most of this treeshakes away during bundling.
  - Cave doesn't actually have sessionStorage, but for some reason it has two localStorages, so put a startup hook in to clear one of them so we can use it as if it was session storage.
- Have pjax-lite save its history into the sessionStorage and load it on setup. This fixes issues with losing the page state across page loads. Various other pjax-lite fixes.
  - Check if we have history on initial load (from storage) and toggle back button.
  - Better handling for XHR errors.
  - Support for only refreshing a part of a page.
  - Support for non-links using data-href (currently only .post-content)
- Move pjax-lite into common and port Portal to use it instead of full-fat PJAX. This means both console frontends are on the same SPA framework, and it's our in-house one that only implements the features we actually need!
  - Obviously this is a very major change to how navigation works on portal and we need to test ~everything.
  - Fixed a bug where pressing the B button for longer than 100-200ms (depends on timing) would go back several times
  - When testing on browser, the native back/forward buttons don't work right anymore since we aren't using the History API. Same as CTR. Use the onscreen ones please!

User changelog could probably be like...
- Fixed 3DS losing browser history and jumping around to random pages when going back
- Fixed Wii U exit button appearing at the wrong times
- Fixed lots of small issues related to navigation

Very proud of this one. Fixing aliases is worth the price of entry alone!


apps/juxtaposition-ui/webfiles/portal/js/components/ui/PortalNavTabs.ts
```diff
@@ -29,10 +30,10 @@ function navTabsClick(this: HTMLElement, ev: Event): void {
 		}
 
 		target.innerHTML = xhr.responseText;
-		window.history.pushState({ url: href, title: '', scrollPos: [0, 0] }, '', href);
+		pjaxSetUrl(href, true);
```
mrjvs, line 33: This makes it so scroll position is no longer reset.
A place where this is an issue is when scrolling down on a feed and clicking on a community, it will not put you at the community header but scrolled down to a random post

apps/juxtaposition-ui/webfiles/portal/js/components/PortalNavBar.ts
```diff
@@ -18,6 +19,24 @@ function backButton(ev: Event): void {
 	back();
 }
 
+function updateBackButton(): void {
```
mrjvs, line 22: The navigation tabs don't re-render when pressing back. That means you can navigate to your notifications, then press the back button. Notifications will be highlighted while you're no longer there.

apps/juxtaposition-ui/webfiles/portal/js/components/PortalNavBar.ts
```diff
@@ -33,4 +52,6 @@ export function initNavBar(): void {
 		navBack.href = '#';
 		navBack.addEventListener('click', backButton);
 	});
+
+	updateBackButton();
```
mrjvs, line 56: Unsure why, but when first loading portal, it shows the go-back button instead of close.

apps/juxtaposition-ui/webfiles/portal/js/components/PortalNavBar.ts
```diff
@@ -18,6 +19,24 @@ function backButton(ev: Event): void {
...
 }
 
+function updateBackButton(): void {
+	var back = document.getElementById('nav-menu-back');
+	var close = document.getElementById('nav-menu-exit');
+	if (!back || !close) {
+		return;
+	}
+
+	if (pjaxCanGoBack()) {
```
mrjvs, line 29: I thought my opinion on this fully through yet but I do want to have it said somewhere for discussion.
With this implementation, a user only ever sees the close button on applet open. Then it will always be a back button.

That feels little much. Maybe only showing the back button when being a layer deeper than the root tabs could work? (although it would be hell to implement)
Idk, im not sure what the best approach is, it just doesn't feel right.

apps/juxtaposition-ui/webfiles/portal/js/juxt.js
```diff
@@ -190,47 +183,29 @@ function initAll() {
...
 document.addEventListener('DOMContentLoaded', function () {
-	pjax = new Pjax({
-		elements: 'a[data-pjax]' +
-			'',
-		selectors: ['title', '#body'],
-		switches: { '#nav-menu': Pjax.switches.replaceNode, '.tab-body': Pjax.switches.replaceNode }
+	pjaxInit({
+		elements: '[data-pjax]',
+		selectors: ['#body']
 	});
```
mrjvs, line 207: Pjax-lite doesn't use real browser navigation. One downside this has is that the real URL never updates.
For example, refreshing the tab on browser when browsing portal will now always put you back to homepage. This can be annoying when debugging.

Idk if it's really a problem, but I would like to keep normal browser functionality working when adopting pjax-lite if possible.
If it keeps working, we can also start adopting pjax-lite into web.

mrjvs (changes requested):
Very good improvements! The new common library also inspired me to start on frontend modules again, I made it a stack with this one: #605 

Just a couple regressions to resolve, hopefully they don't take too much time to fix.

mrjvs:
Oh also I updated the tsconfigs a little so there's less copy-pasting

apps/juxtaposition-ui/webfiles/portal/js/components/PortalNavBar.ts
ashquarky: Can't reproduce this
image

Remember pjax-lite stores history in sessionStorage now, so if you reload or navigate around from a browser it will remember the old history. You need to fully close and re-open the tab to get a fresh state.

This is a regression in browser debugging, but it very cleanly solves the pjax hydration issues on consoles (which always wipe the sessionStorage anyway) so I figured it was an okay tradeoff

apps/juxtaposition-ui/webfiles/portal/js/juxt.js
ashquarky: Yeah, this is how debugging on CTR has been for a little while, it is a little unfortunate. Fixing it would mean including a bunch of dead code in the bundles that will never get used on console, which I'm not sure how I feel about.

apps/juxtaposition-ui/webfiles/portal/js/components/PortalNavBar.ts
ashquarky: I don't disagree, however this is how it's worked historically so I think for this PR it's fine to just keep the behaviour the same

apps/juxtaposition-ui/webfiles/portal/js/components/PortalNavBar.ts
ashquarky: Believe it or not this is how it works today? I guess we must have broken it at some point. I intend to submit another PR to redo how this works, a-la data-toolbar on CTR

apps/juxtaposition-ui/webfiles/portal/js/components/ui/PortalNavTabs.ts
ashquarky: Good catch

apps/juxtaposition-ui/webfiles/portal/js/components/PortalNavBar.ts
ashquarky: Originally I did have it swap the toolbar too which is how I found out that the SSR highlighting for userpage and feed is totally broken, and a lot of pages don't have any highlight set, so. Yeahh

apps/juxtaposition-ui/webfiles/portal/js/juxt.js
ashquarky: *Fixing it for the debugging case, I mean. For a hypothetical web release we can use platform checks like storage does

PR #605, opened 2026-09-16
Frontend modules: 2 electric boogaloo

mrjvs:
Changes:
- Implemented frontend modules (from #424) into common library, now called `islands` because it quite closely resembles the islands frontend pattern. (and I didn't like the name "module" since its not very specific)
- Added README about islands, in case any contributors are curious about what it means.
- Cleaned up eslint config:
  - Put all webfiles rules into `webfiles/base.eslint.config.mjs`, makes it easier to work work
  - Add rules to restrict unsupported JS syntax on JS files that target consoles
  - Disabled eslint on polyfills
- Ported some CTR JS to islands: `controls/ctabs` and `controls/checkbox.ts`

I will be porting the rest of the JS code to islands after this is merged. That way there is less to test upfront and I can split up the refactoring into smaller PRs.

This will need some testing with a real 3DS console or emulator. I've tested my changes in browser.


apps/juxtaposition-ui/webfiles/common/js/islands/island-container.ts
```diff
@@ -0,0 +1,52 @@
...
+export type IslandContainer = {
+	register: (mod: IslandArray) => void;
+	loadBody: () => void;
+	loadPartial: (el: HTMLElement) => void;
+};
+
+export function createIslandContainer(ops?: IslandContainerOptions): IslandContainer {
+	var islands: IslandControls[] = [];
+	function registerManyIslands(islandArr: IslandArray): void {
+		for (var i = 0; i < islandArr.length; i++) {
```
ashquarky: Pretty sure Array.concat is available everywhere

apps/juxtaposition-ui/webfiles/common/js/islands/island-container.ts
```diff
@@ -0,0 +1,52 @@
...
+			var hasDoubleHydrated = false;
+
+			var ctx: IslandMountHookContext = {
+				doc: el,
+				triggerDoubleHydrate: () => hasDoubleHydrated = true
+			};
+			for (var i = 0; i < islands.length; i++) {
+				var island = islands[i];
+				island.run(ctx);
+			}
```
ashquarky: Array.forEach is also polyfilled

ashquarky (dismissed):
Still need to test it but i like the design. A lot lighter than the old one and seems to cover the needs we have

apps/juxtaposition-ui/webfiles/common/js/islands/island-container.ts
```diff
@@ -0,0 +1,52 @@
...
+
+			if (hasDoubleHydrated) {
+				var text = 'Double hydration detected, have you called loadPartial twice?';
```
ashquarky, line 43: It's worth noting that iirc all of today's handlers are written to be resistant to double hydration.. but still good to have some dev-checks for it

mrjvs: its a pretty cheap check to do, so I think it would be fine to keep it. Since it warns and does a popup, it's gonna be very obvious if we ever do double-hydration and can avoid it. That way we don't have to make our handlers double-hydration-compatible anymore. (less to think about is always better)

apps/juxtaposition-ui/webfiles/common/js/islands/island-container.ts
mrjvs: Fixed

apps/juxtaposition-ui/webfiles/common/js/islands/island-container.ts
mrjvs: Fixed

ashquarky (approved):
Works on 3DS 👍

PR #608, opened 2026-09-17
Experimental: Markdown styling in post content

mrjvs:
A part of #67 

Changes:
- Introduce common library that handles markdown parsing and rendering
- Added unit tests to common library
- Added full markdown flow: transformation -> parsing -> rendering
  - **Transformation:** Turn shorthands into easily parsable identifiers. For example we can add `@pnid` which turns into `<@pid>`. There are currently no transformations yet.
  - **Parsing:** Turn markdown input string into an AST
  - **Rendering:** Render the AST into plain text or html
- Added one extra markdown plugin to preserve linebreaks, otherwise they all just disappear with the markdown parsing.
- Added `<PostContent />` component to render a post in either plain text or markdown.
- Added markdown to new posts. Text version gets stored in the normal body field. Markdown gets stored in a new db field.
- Add unit tests to CI

Will still need to be discussed with moderation team and dev team to determine if it's a good idea or not


mrjvs:
It's ready for review, but will still need to be discussed with moderation team team and dev team to determine if it's viable.

It has some clear downsides, so it's worth discussing before merge.

